PR #9, opened 2018-10-12
download_community improvements

fbartels:
fix shellcheck warnings on download_community
remove gzip and chmod from dockerfile to script itself

That is a nice trick with the h5ai api. in the past I have used a combination of curl and lynx to pick the right archive, but parsing the json response of h5ai is much nicer.


webapp/Dockerfile
```diff
@@ -43,12 +43,10 @@ COPY download_community.sh /kopano/download_community.sh
 
 # install kopano web app and refresh ca-certificates
 RUN \
-    # community download 
+    # community download
     [ ${DOWNLOAD_COMMUNITY_PACKAGES} -eq 1 ] && \
-    chmod a+x /kopano/download_community.sh && \
     /kopano/download_community.sh core && \
     /kopano/download_community.sh webapp && \
-    gzip -f9 Packages; \
```
zokradonh, line 51: Please note that this line ends with `;`. So it finishes the fake "if"-statement on DOWNLOAD_COMMUNITY_PACKAGES
`/kopano/download_community.sh webapp && \` should be
`/kopano/download_community.sh webapp; \`

zokradonh:
> 100644 → 100755

Ah nice, I didn't know that git is capable of this.

>fix shellcheck warnings on download_community`

Oh yes thank you!

>That is a nice trick with the h5ai api. in the past I have used a combination of curl and lynx to pick the right archive, but parsing the json response of h5ai is much nicer.

Yes I also like h5ai much more, but it may brake if you change your distribution server software.

I will merge your changes as soon as you finished.

PR #20, opened 2018-11-11
Invalid command ending

zokradonh:


fbartels:
Oh, good catch. I should have tried to rebuild the image before making the z-push pr.

zokradonh:
I was on the rush and saw the error message in the docker hub build log :)

PR #40, opened 2018-11-30
rework current front facing proxy

fbartels:
this replaced the current approach to caddy with a custom build of kweb and requires minimal configuration once checked out.

@zokradonh can you add an automated build for `kweb` and `ldap-demo`?


Makefile
```diff
@@ -53,6 +53,9 @@ build-zpush:
 build-ssl:
 	docker build -t $(docker_repo)/kopano_ssl ssl/
 
+build-kweb:
+	docker build --no-cache -t $(docker_repo)/kopano_web kweb/
```
longsleep: `--no-cache` intentional?

docker-compose.yml-example
```diff
@@ -2,21 +2,20 @@ version: "3"
...
     restart: always
-    privileged: true
     links:
       - kopano_webapp
       - kopano_zpush
     ports:
       - "2015:2015"
       - "${HTTP}:80"
       - "${HTTPS}:443"
+    command: caddy -conf /etc/Caddyfile -agree -email ${EMAIL} -host ${FQDN}
```
longsleep: Rename /etc/Caddyfile to /etc/kweb.cfg or something.

longsleep: This needs checking if that start up fine when fqdn is given. Probably needs checks to avoid those parameters when it the variables are not given (needs a docker-run script).

kweb/Caddyfile
```diff
@@ -0,0 +1,30 @@
+*, http://, https:// {
+    gzip
+
+    tls self_signed
```
longsleep: the `self_signed` value should be coming from the environment (use the email, or if no host is set, use self_signed)

kweb/Caddyfile
```diff
@@ -0,0 +1,30 @@
...
+
+    redir {
+        if {scheme} is http
+        / https://{host}{uri}
+    }
```
longsleep: This should be put into the `http://` vhost in a simplified way.

kweb/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM kopano/kwebd:0.4.2
+RUN apk add --no-cache libcap && \
+    setcap 'cap_net_bind_service=+ep' /usr/local/bin/kwebd
```
longsleep: It would be better if we would only listen on two ports (like 8080 and 8443) which automatically does not require that special cap.

docker-compose.yml-example
fbartels: renamed and yes, starts up with these parameters (well at least if the env variables do exist)

zokradonh:
> @zokradonh can you add an automated build for `kweb` and `ldap-demo`?

zokradonh/kopano_web
zokradonh/kopano_ldap_demo
Automatically built on push to zokradonh/kopano-docker.git

README.md
```diff
@@ -55,6 +52,13 @@ If you are running a private Docker Registry then you may also change `docker_re
 
 The built image includes your subscription key! Do not push this image to any public registry like e.g. https://hub.docker.com!
 
+### What if I want to use a different front facing proxy than the one in docker-compose? Or just some part of the compose file?
+
+While using kweb is recommended, this is of course possible.
+
+- The `kopano_webapp` image is accessible on port 80 and serves the WebApp both on `/` and `/webapp`.
+- The `kopano_zpush` image is accessible on port 80 and servers Z-Push on `/Microsoft-Server-ActiveSync` (additional urls may be needed in the future see #39).
```
zokradonh:
```suggestion
- The `kopano_zpush` image is accessible on port 80 and serves Z-Push on `/Microsoft-Server-ActiveSync` (additional urls may be needed in the future see #39).
```
Typo

docker-compose.yml-example
```diff
@@ -2,21 +2,20 @@ version: "3"
 
 services:
   web:
-    build: caddy/
-    #image: abiosoft/caddy:0.10.4
+    build: kweb/
```
zokradonh: `build` not required anymore. We can use `image`

zokradonh (approved):

fbartels:
@longsleep with the last commit I have introduced a small and crude wrapper script. this at least catches the case to not set -host if the email is set as 'self_signed'.

nupplaphil:
Great work! Getting rid of the `docker.socket` dependency was a great step further!

Issue #42, opened 2018-12-01
simplify setup.sh


fbartels:
most options may be too confusing for some users.
- get fqdn from ´hostname -f´
- get rid of usernames and passwords to be entered
- get rid of ldap questions
- ... ?

fbartels:
Resolved by #43

PR #43, opened 2018-12-02
Simplifying setup.sh

nupplaphil:
FQDN is now used to create default base DN and default domain for postmaster.

Part of #42

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -3,7 +3,11 @@
 set -e
 
 fqdn_to_dn() {
-    printf 'dc=%s' "$1" | sed -r 's/\./,dc=/g'
+	printf 'dc=%s' "$1" | sed -r 's/\./,dc=/g'
+}
+
+random_string() {
+	hexdump -n 16 -v -e '/1 "%02X"' /dev/urandom
 }
 
 if [ ! -e ./docker-compose.yml ]; then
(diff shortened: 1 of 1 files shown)


nupplaphil:
- The guessed base DN is now based on the FQDN
- The timezone is guessed based on the local system
- If the default LDAP (= bundled) server is used, the PW is now random
- If the default DB (= bundeld) is used, the PW is now random

fbartels:
Thanks for your pr. I will need a bit of time to go through the changes, since I am on holiday until Saturday and do not have a full system to play with it. Will check back next weekend.

setup.sh
```diff
@@ -24,63 +32,86 @@ if [ ! -e ./.env ]; then
...
+		value_default="CN=readonly,$LDAP_BASE_DN"
+		read -p "LDAP bind user (needs only read permissions) [$value_default]: " new_value
+		LDAP_BIND_DN=${new_value:-$value_default}
+
+		value_default="kopano123"
+		read -p "LDAP bind password to be used [$value_default]: " new_value
+		LDAP_BIND_PW=${new_value:-$value_default}
+	else
+		LDAP_ADMIN_PASSWORD=$(random_string)
+		LDAP_SEARCH_BASE="$LDAP_BIND_DN"
```
fbartels: when just entering the setup, the ldap_search_base is empty for me.

setup.sh
```diff
@@ -24,63 +32,86 @@ if [ ! -e ./.env ]; then
 	value_default="kopano.demo"
 	read -p "FQDN to be used (for reverse proxy) [$value_default]: " new_value
 	FQDN=${new_value:-$value_default}
+	LDAP_BASE_DN=$(fqdn_to_dn $FQDN)
```
fbartels: move this down just above line 41?

setup.sh
fbartels: it should be LDAP_BASE_DN instead of LDAP_BIND_DN

setup.sh
```diff
@@ -24,63 +32,86 @@ if [ ! -e ./.env ]; then
...
-	value_default="CN=readonly,DC=kopano,DC=demo"
-	read -p "LDAP bind user (needs only read permissions) [$value_default]: " new_value
-	LDAP_BIND_DN=${new_value:-$value_default}
-
-	value_default="kopano123"
-	read -p "LDAP server to be used (default bundled openldap) [$value_default]: " new_value
-	LDAP_BIND_PW=${new_value:-$value_default}
+	if [ "$LDAP_SERVER" != "$value_default" ]; then
+		value_default="kopano123"
+		read -p "Password of the admin user [$value_default]: " new_value
```
fbartels: When not using the bundled ldap then we do not have to differentiate between  admin/read-only user. remove the ldap_admin_password option.

setup.sh
```diff
@@ -24,63 +32,86 @@ if [ ! -e ./.env ]; then
...
-	read -p "Password to connect to the database [$value_default]: " new_value
-	MYSQL_PASSWORD=${new_value:-$value_default}
-
-	value_default="kopano"
-	read -p "Datebase to use for Kopano [$value_default]: " new_value
-	MYSQL_DATABASE=${new_value:-$value_default}
-
+	if [ "$MYSQL_HOST" != "$value_default" ]; then
+		value_default="kopano123"
+		read -p "Password for the MySQL root user [$value_default]: " new_value
```
fbartels: here its the same as with the ldap image. root user not needed, we just need a user to query the database (and potentially create tables)

fbartels:
Maybe at the end of setup.sh print something like "hey, you specified a different ldap/db than what is bundled with compose. Remember to remove these from the compose file"

fbartels (changes requested):
obligatory comment to mark this pr as "request changes"

nupplaphil:
I had to rebase because of missing changes in the setup

fbartels (approved):

fbartels:
Thanks looks good now. I saw a smaller thing in the casing of the ldap dns, will fix that after merge myself.

PR #44, opened 2018-12-02
LDAP demo based on setup

nupplaphil:
Using LDAP environment variables instead of fixed "kopano.demo" for ldap-demo docker image


fbartels:
Thanks for the pr!

PR #49, opened 2018-12-09
Adding additional repositories to WebApp setup

nupplaphil:
I believe that security is important.
So I'd like to generally add the S/MIME support to the webapp image.


fbartels:
Hi, I like the idea. Given that there are some other plugins for webapp that could also be added I'll make some comments inline.

base/Dockerfile
```diff
@@ -41,6 +41,8 @@ ARG KOPANO_CORE_VERSION=newest
 ARG KOPANO_REPOSITORY_FLAGS="trusted=yes"
 ARG KOPANO_WEBAPP_REPOSITORY_URL="file:/kopano/repo/webapp"
 ARG KOPANO_WEBAPP_VERSION=newest
+ARG KOPANO_SMIME_REPOSITORY_URL="file:/kopano/repo/smime"
```
fbartels: Since there is also a separate repo/download for mdm and files I would suggest to rather name this variable `KOPANO_WEBAPP_SMIME_REPOSITORY_URL`

webapp/Dockerfile
```diff
@@ -66,6 +71,7 @@ RUN \
         kopano-webapp-plugin-titlecounter \
         kopano-webapp-plugin-webappmanual \
         kopano-webapp-plugin-zdeveloper \
+        kopano-webapp-plugin-smime \
```
fbartels: the rest of the packages in the dockerfile are sorted alphabetically

fbartels (changes requested):
The variable for the s/mime repo should also be added to setup.sh (commented as the other ones). The correct url for this is https://download.kopano.io/supported/smime:/final/Debian_9.0/.

I don't think that there are changes needed to the makefile, though.

nupplaphil:
What about adding the repository for mdm / files as well?

So we can add them to the setup.sh like "Do you want to enable MDM support?" and "Do you want to enable Files support?"

fbartels:
Yes, I think in the end at least the repositories for all these plugins should be configured, so users can easily install them if necessary.

One thing to be aware of is that all installed plugins are loaded for the user when he logs into webapp, so ideally you do not want to have too many (unused) plugins installed. So in addition to this the current list of plugins should be reduced as well (no contactfax, no zdeveloper, less spell check plugins).

But you are right, we could include a question in setup.sh that then adds the according packages to `ADDITIONAL_KOPANO_PACKAGES`.

fbartels:
@nupplaphil can you rebase? 

One minor style "complaint". The setup.sh script is using tabs, but in your last pr you already mixed in spaces and here you are using even more of them.

fbartels:
Hi again,

I tried your changes this morning in a direct checkout of your branch (opposed to pulling in your branch into a locally created branch, which suffered from merge conflicts). I quite like your changes.

Some small notes:
- https://github.com/nupplaphil/kopano-docker/blob/smime/setup.sh#L11 I would sort this list alphabetically.
- the only way to get plugins in WebApp is now to specify them in ADDITIONAL_KOPANO_WEBAPP_PLUGINS. One one side I like that WebApp is now as minimal as possible, which should reduce login times. On the other hand I think there should be a small set of default plugin (recommendations) like the desktop notifications or the title counter. I think in the end we have to see what other users think of this and maybe adapt this later on again. https://stash.z-hub.io/projects/K4U/repos/kopano4ucs/browse/debian/control#99-112 are btw the plugins that are installed by default on Univention.

nupplaphil:
I'll hopefully add tomorrow the suggested improvements :-) sounds good for me!

fbartels:
Just to be clear. For the moment you do not have to change anything about the by default installed packages.

fbartels:
Totally missed that you update this already yesterday. Looks good now merging it.

Thanks for your work!

Issue #60, opened 2018-12-20
save search index in volume


nupplaphil:
I had to regenerate my containers and therefore lost the search index.

This seems to cause z-push errors like https://forum.kopano.io/topic/1678/statusexception-exportchangesics-initializeexporter-error-mapi_exportchanges_config-failed-0xffffffff8004010f

After some time, the folder reappears on the devices again when I'm manually forcing the folder to get indexed (f.e. adding an item in the webapp or receiving a email)

So I'd say we should save the index state into a own volume.

fbartels:
Yes, that is indeed a good idea. I would propose to set the values of `index_path` in https://github.com/zokradonh/kopano-docker/blob/master/core/defaultconfigs/search.py#L8. So that data is by default also kept in `/kopano/data`. e.g. in `/kopano/data/search`. 

```
# Location of the index files
#index_path = /var/lib/kopano/search/
```

nupplaphil:
I think I'll add it soon, because the whole server went mad ;-) .. Have to resync all devices (but after the index is finished..)

nupplaphil:
Fixed with #61

PR #61, opened 2018-12-20
Adding search index to data container

nupplaphil:
Fixing #60 

[Edit] I already switched with my docker environment and i works flawless with `/kopano/data/search/` :-)


fbartels:
Thanks!

PR #65, opened 2019-01-02
ldap network fix and improvmements

cguentherTUChemnitz:
see commit messages:
 - fix ldap admin does not reach the ldap service
 - use statically named ldap-net to be able to connect external services more easily
 - split kopano networking into kopanonet and ldap-net


docker-compose.yml-example
```diff
@@ -320,3 +323,6 @@ networks:
   web:
   kopanonet:
     driver: bridge
+  ldap-net:
+    name: ldap-net
```
fbartels, line 327: Since this network has a dash in its name, why not rename the other network to have a dash as well? Instead of "web" then also "public-web"?

docker-compose.yml-example
```diff
@@ -1,4 +1,4 @@
-version: "3"
+version: "3.5"
```
fbartels, line 1: Is there a special reason for increasing the version? (just curious)

README.md
```diff
@@ -13,6 +13,7 @@ This repository contains an easy to replicate recipe to spin up a [Kopano](https
 - now run `docker-compose up` and you will see how the remaining Docker images are pulled and then everything is started
 - after startup has succeeded you can access the Kopano WebApp by going to `https://kopano.demo/webapp`
 - you can also access phpLDAPadmin by going to `https://kopano.demo/ldap-admin`
+  - you can access the ldap admin with: `dc=admin,dn=kopano,dn=demo` as login dn and the `LDAP_ADMIN_PASSWORD` environment variable provided by the .env file (which is generated by the setup.sh)
```
fbartels: The username is also dynamic. Maybe also tell users to read it from the .env file?

fbartels (changes requested):
Thanks for your pr. I have added some remarks inline.

docker-compose.yml-example
cguentherTUChemnitz: The network name entry is introduced in compose version 3.5
https://docs.docker.com/compose/compose-file/#name-1

docker-compose.yml-example
cguentherTUChemnitz: Good point. I will update those.

README.md
cguentherTUChemnitz: I did not find the ldap admin username in the env. file. I did also not find it in the  ldap-admin environment setup of the compose file. So i luckily estimated the username based from other setups. Are you sure, that the ldap-admin username is dynamic?

fbartels: Yes, you're right. It is indeed not totally clear just from looking at compose or env. The first part (so `dc=admin,` is given), but the rest is just your ${LDAP_BASE_DN}

docker-compose.yml-example
cguentherTUChemnitz: I updated the exisiting network names for better consistency. I will describe some motivation here to use the name property for only one network. Docker-compose generates network names using an aggregation of the folder-name where the docker-compose.yml lives in and the network name specified in the compose file. So for example the kopano-net will be created as kopano-docker-kopano-net, which emphasizes the affiliation to the docker-kopano service stack. So typically i use those service-stack prefixed networks, when those should be used only by services inside this service stack. When a service should be used also by serivces of other service stacks, like a proxy or ldap, i ensure non-prefixed network names to highlight the service stack overlapping behavior and easier name deduction in other compose files.

If my explanation and the more consistent network names fits your expectations, you can please resolve this conversation :)

README.md
cguentherTUChemnitz: Oh jeah, you are totally right. I will add some more hints.

cguentherTUChemnitz: Now the readme is more specific about reading the login information from the .env file

docker-compose.yml-example
fbartels: Thanks

README.md
fbartels: ah, I see I mixed up it up here as well, its cn=admin. I'll quickly fix this up and then merge this pr.

Thanks for your work!

fbartels (approved):

fbartels:
manually merged in https://github.com/zokradonh/kopano-docker/commit/a9ef309cb09ed888cda5d8f5a121c227fbfed6eb

Issue #66, opened 2019-01-02
how to recreate container with persistent data?


cguentherTUChemnitz:
I am testing the proposed compose setup. Some services are not able to startup properly after restart or recreation. What i did:
 - running setup.sh with default settings
 - starting containers via docker-compose up -d
 - minimal config change in docker-compose, enforcing to recreate the container
 - recreate container via docker-compose up -d

kopano_spooler and kopano_monitor are not able to start again. I assume that the container recreation keeps pids in the volumes, which disturbs their restart process. 

The log of those containers are:
```
Jan 02 19:47:50 Configure core service 'spooler'
Jan 02 19:47:50 Set ownership
2019/01/02 19:47:50 Waiting for: file://var/run/kopano/server.sock
2019/01/02 19:47:50 Waiting for: tcp://mail:25
2019/01/02 19:47:50 Problem with dial: dial tcp 172.24.0.4:25: connect: connection refused. Sleeping 1s
2019/01/02 19:47:51 File file://var/run/kopano/server.sock had been generated
2019/01/02 19:47:51 Problem with dial: dial tcp 172.24.0.4:25: connect: connection refused. Sleeping 1s
2019/01/02 19:47:52 Problem with dial: dial tcp 172.24.0.4:25: connect: connection refused. Sleeping 1s
2019/01/02 19:47:53 Problem with dial: dial tcp 172.24.0.4:25: connect: connection refused. Sleeping 1s
2019/01/02 19:47:54 Problem with dial: dial tcp 172.24.0.4:25: connect: connection refused. Sleeping 1s
2019/01/02 19:47:55 Problem with dial: dial tcp 172.24.0.4:25: connect: connection refused. Sleeping 1s
2019/01/02 19:47:56 Problem with dial: dial tcp 172.24.0.4:25: connect: connection refused. Sleeping 1s
2019/01/02 19:47:57 Problem with dial: dial tcp 172.24.0.4:25: connect: connection refused. Sleeping 1s
2019/01/02 19:47:58 Connected to tcp://mail:25
Wed Jan  2 19:47:58 2019: [crit   ] Warning: Process /usr/sbin/kopano-spooler is probably already running.
Wed Jan  2 19:47:58 2019: [crit   ] If you are sure the process is stopped, please remove pidfile /var/run/kopano/spooler.pid
Wed Jan  2 19:47:58 2019: [crit   ] main(): Failed creating PID file
Jan 02 19:51:39 Configure core service 'spooler'
[...]

PR #68, opened 2019-01-02
fix startup of core-image containers after unclean shutdown

cguentherTUChemnitz:
as discussed here:
https://github.com/zokradonh/kopano-docker/issues/66

### DIFF
--- a/core/start-service.sh
+++ b/core/start-service.sh
@@ -8,6 +8,9 @@ if [ ! -e /kopano/$SERVICE_TO_START.py ]
 then
     echo "Invalid service specified: $SERVICE_TO_START" | ts
     exit 1
+else
+	#ensure removed pid-file on unclean shutdowns and mounted volumes
+	rm -f /var/run/kopano/$SERVICE_TO_START.pid
 fi
 
 [ ! -z "$ADDITIONAL_KOPANO_PACKAGES" ] && apt update


core/start-service.sh
```diff
@@ -8,6 +8,9 @@ if [ ! -e /kopano/$SERVICE_TO_START.py ]
 then
     echo "Invalid service specified: $SERVICE_TO_START" | ts
     exit 1
+else
+	#ensure removed pid-file on unclean shutdowns and mounted volumes
+	rm -f /var/run/kopano/$SERVICE_TO_START.pid
```
fbartels, line 13: I did not try this yet, but the bash script will immediately exit on non zero return codes and if you put it into the else the command will always be executed (and definitely fail on a fresh start).

I would recommend to put it in its own if and check if the file exists before deleting it. Forcing rm also seems drastic .

fbartels (changes requested):

core/start-service.sh
zokradonh: Just tested this.  `rm` on non existent files return non zero return code, yes. But if you provide `-f` it always returns zero return code. :)

fbartels: ok, then this change at least works as intended.

What do you think @zokradonh keep it in the else or make an own check for it? (its more of a style question, then)

cguentherTUChemnitz: Maybe this is interesting for the discussion. I tried to keep it in sync with the style of the z-push apache pid file removal on startup:
https://github.com/zokradonh/kopano-docker/blob/87c814d491a88b442ee98d3100786a10fb111f74/zpush/start.sh#L43

fbartels: Huh, indeed.

For that case there is no variable involved, though. So what if:
```
SERVICE_TO_START="../../ "
rm -f /var/run/kopano/$SERVICE_TO_START.pid
```
But that is all hypothetical.

cguentherTUChemnitz: Ok i see your point. I think to get more robustness, we could check that the expression /var/run/kopano/$SERVICE_TO_START.pid points at least to a file with ending .pid in the desired folder.

But keep in mind, that no -r flag is provided. So as far as i know it would not be able to delete folders anyway.

cguentherTUChemnitz: > Ok i see your point. I think to get more robustness, we could check that the expression /var/run/kopano/$SERVICE_TO_START.pid points at least to a file with ending .pid in the desired folder.

I do not have much experience in writing non-exploitable bash checks. So if you want more checks here, please add those as you like.

fbartels: Yes, valid point about not having the -r flag. I think I have "gepiesackt" you for long enough. ;-) Lets keep the `rm -f` for now. I will though still make a small styling change after merging this.

Thanks for your contribution!

PR #70, opened 2019-01-04
initial setup of an self-service

cguentherTUChemnitz:
TODO:
 - [x] update container tag to statically versioned service

I add a preconfigured password self service as additional folder-based web service, which i tried to integrate seamless with the ldap, mail stack and existing reverse proxy.

Use this pull request as an proposal and a start-point for discussion ;)


docker-compose.yml-example
```diff
@@ -55,6 +55,43 @@ services:
       - ldap-net
       - web-net
 
+  password-self-service:
+    image: tiredofit/self-service-password
```
fbartels: I would recommend to always use a fixed tag

docker-compose.yml-example
```diff
@@ -55,6 +55,43 @@ services:
...
+    container_name: password-self-service
+    domainname: ${LDAP_DOMAIN}
+    depends_on:
+      - ldap
+      - mail
+    environment:
+      - LDAP_SERVER=ldap://ldap:389
+      - LDAP_BINDDN=cn=admin,dc=kopano,dc=demo
```
fbartels: should use cn=admin,${LDAP_BASE_DN} here

setup.sh
```diff
@@ -209,6 +209,15 @@ LDAP_QUERY_FILTER_ALIAS=(&(kopanoAccount=1)(kopanoAliases=%s))
 LDAP_QUERY_FILTER_DOMAIN=(&(|(mail=*@%s)(kopanoAliases=*@%s)))
 SASLAUTHD_LDAP_FILTER=(&(kopanoAccount=1)(uid=%s))
 
+# LDAP user password self-service reset settings
+SELF_SERVICE_SECRETEKEY=$(random_string)
+SELF_SERVICE_PASSWORD_MIN_LENGTH=
```
fbartels: predefine some policies here?

fbartels (commented):
Hi,
I like the idea and ssp in general (I actually quite a few years back contributed the german translation, although I don't remember if I started from scratch or just completed what was already there).

In general the image looks nice, since it allows a lot of configuration through env variables, the one thing I do not like about it is that it just does a checkout of the ssp master branch on first start instead of using a proper release. this makes this docker image not really reproducible. 

(see https://github.com/tiredofit/docker-self-service-password/blob/master/install/etc/s6/services/10-nginx/run#L82-L89)

ps: if you add the email that you use in your .git/config to your github profile you would be shown as a contributor of this project and not as a 'first time contributor' in this pr.

docker-compose.yml-example
```diff
@@ -55,6 +55,43 @@ services:
...
+      - PASSWORD_MIN_LENGTH=${SELF_SERVICE_PASSWORD_MIN_LENGTH}
+      - PASSWORD_MAX_LENGTH=${SELF_SERVICE_PASSWORD_MAX_LENGTH}
+      - PASSWORD_MIN_LOWERCASE=${SELF_SERVICE_PASSWORD_MIN_LOWERCASE}
+      - PASSWORD_MIN_UPPERCASE=${SELF_SERVICE_PASSWORD_MIN_UPPERCASE}
+      - PASSWORD_MIN_DIGIT=${SELF_SERVICE_PASSWORD_MIN_DIGIT}
+      - PASSWORD_MIN_SPECIAL=${SELF_SERVICE_PASSWORD_MIN_SPECIAL}
+    expose:
+      - "80"
+    volumes:
+      - password-self-service:/www/ssp
```
fbartels: in relation to the comment with doing a git checkout of ssp instead of using a release. having this as a proper volume means that users would need to enter into the running container and doing a `git pull` to get a newer ssp version. if this volume is removed it would download a new release once the container is recreated locally.

cguentherTUChemnitz:
Thanks for your productive review. I will connect with the container maintainer trying to avoid the pitfalls.

docker-compose.yml-example
cguentherTUChemnitz: I started a discussion to solve this at the used docker image project:
https://github.com/tiredofit/docker-self-service-password/issues/10

setup.sh
cguentherTUChemnitz: Which default policies would you like to see?

docker-compose.yml-example
fbartels: to have it said here as well, afais just remove the volumes section from this image

docker-compose.yml-example
```diff
@@ -309,6 +346,7 @@ services:
 volumes:
   web:
   ldap:
+  password-self-service:
```
fbartels: and remove this line as well

cguentherTUChemnitz:
Please do not merge this current state of the PR. The currently referenced container tag 2.8 is not running properly, because it misses the variable replacement fix:
https://github.com/tiredofit/docker-self-service-password/pull/8

So in my eyes, we can merge after @tiredofit creates a static labeled image based on the discussion in:
https://github.com/tiredofit/docker-self-service-password/issues/10

setup.sh
fbartels: Did miss this one before. Good question. The project itself defines it like this https://github.com/ltb-project/self-service-password/blob/master/conf/config.inc.php#L91-L105

I would say lets put all zeros there by default. having a number in there should at least make it easier for admins to see that this is something they could/should customise.

docker-compose.yml-example
```diff
@@ -55,6 +55,41 @@ services:
...
+      - PASSWORD_NO_REUSE=true
+      - WHO_CAN_CHANGE_PASSWORD=user
+      - SECRETEKEY=${SELF_SERVICE_SECRETEKEY}
+      - BACKGROUND=.
+      - PASSWORD_MIN_LENGTH=${SELF_SERVICE_PASSWORD_MIN_LENGTH}
+      - PASSWORD_MAX_LENGTH=${SELF_SERVICE_PASSWORD_MAX_LENGTH}
+      - PASSWORD_MIN_LOWERCASE=${SELF_SERVICE_PASSWORD_MIN_LOWERCASE}
+      - PASSWORD_MIN_UPPERCASE=${SELF_SERVICE_PASSWORD_MIN_UPPERCASE}
+      - PASSWORD_MIN_DIGIT=${SELF_SERVICE_PASSWORD_MIN_DIGIT}
+      - PASSWORD_MIN_SPECIAL=${SELF_SERVICE_PASSWORD_MIN_SPECIAL}
```
longsleep, line 86: This should make sure that PASSWORD_HASH is something useful, from the docs at https://github.com/tiredofit/docker-self-service-password clear seams to be the default which is brr ..

So for best security this should use {CRYPT} which leaves security to the underlaying os. Not sure if and how the underlaying scripting supports creating of the proper CRYPT salt formatting and what consequence it has with alpine - needs investigation but if we decide to ship as service which can set ldap passwords, it is required that the passwords are stored in a secure fashion. See http://manpages.ubuntu.com/manpages/bionic/man3/crypt.3.html for a useful glibc crypt implementation with formatting rules for hashing alg and rounds. Useful would be SHA-512 with a 16 characters random salt read from urandom and something like 50000 rounds.

setup.sh
```diff
@@ -209,6 +209,15 @@ LDAP_QUERY_FILTER_ALIAS=(&(kopanoAccount=1)(kopanoAliases=%s))
...
 SASLAUTHD_LDAP_FILTER=(&(kopanoAccount=1)(uid=%s))
 
+# LDAP user password self-service reset settings
+SELF_SERVICE_SECRETEKEY=$(random_string)
+SELF_SERVICE_PASSWORD_MIN_LENGTH=0
+SELF_SERVICE_PASSWORD_MAX_LENGTH=0
+SELF_SERVICE_PASSWORD_MIN_LOWERCASE=0
+SELF_SERVICE_PASSWORD_MIN_UPPERCASE=0
+SELF_SERVICE_PASSWORD_MIN_DIGIT=0
+SELF_SERVICE_PASSWORD_MIN_SPECIAL=0
```
longsleep: These seem to be bad defaults. Can we use something at least somewhat secure as default?

longsleep (changes requested):

docker-compose.yml-example
fbartels: Good point. I thought I had seen somewhere that the docker image actually sets PASSWORD_HASH to auto, but this must have been somewhere else.

cguentherTUChemnitz:
A fixed version of the self-service-docker image is used now. This one introduces a new environment variable, which is used to specify a desired release to download. So for the current configuration, we finally get a fixed version of ssp, defined by the docker-compose.yml-example. So in my eyes, this one is now able to merge.

setup.sh
cguentherTUChemnitz: @longsleep Which values would you prefer as a secure default?

setup.sh
cguentherTUChemnitz: The discussion of using proppery defaults is continued here:
https://github.com/zokradonh/kopano-docker/pull/70#discussion_r245516742

docker-compose.yml-example
cguentherTUChemnitz: I my assumption correctly, that this code snipped is the real talking point? Or should we devide here between generating passwords and generating password-hashes?
https://github.com/zokradonh/kopano-docker/blob/246638297e883dad4cdf52fec742a9f6a539177a/setup.sh#L9-L10

setup.sh
longsleep: > Which values would you prefer as a secure default?

Good question - i would say PASSWORD_MIN_LENGTH=5 at the very least and maybe SELF_SERVICE_PASSWORD_MIN_DIGIT=1 with SELF_SERVICE_PASSWORD_MIN_SPECIAL=1 - that should not be too annoying while still preventing stuff like `123` or `password` :)

docker-compose.yml-example
longsleep: > I my assumption correctly, that this code snipped is the real talking point?

No, those are the LDAP passwords to bin ldap and is fine. I am talking about how the individual users passwords are stored in ldap (`PASSWORD_HASH` option).

longsleep:
> So in my eyes, this one is now able to merge.

Please do no merge until the user passwords in LDAP are stored securely.

docker-compose.yml-example
cguentherTUChemnitz: Thanks for clarification. I have now set the PASSWORD_HASH option to crypt. As you mentioned, the alpine based implementation is now used. I don't know what to count on in this musl-libc based distro. Nevertheless, the hashing itself does work. I have testet to reset my password and the ldap-admin now shows crypt as hash type and i am still able to login into the webapp with the new hashed password.

@longsleep thanks for pointing this out. Is this configuration sufficient for a merge?

fbartels (approved):

longsleep:
Not sure how this works now with crypt since form the musl library
> Additional files libm.a, librt.a, libpthread.a, libcrypt.a, libutil.a,
   libxnet.a, libresolv.a, and libdl.a are provided in $(libdir) as empty
   library archives. They contain no code, but are present to satisfy the
   POSIX requirement that options of the form -lm, -lpthread, etc. be
   accepted by the c99 compiler.

longsleep:
This was merged a bit early since crypt without any  options is not secure at all.

longsleep:
I checked and the service behavior depends on php crypt - the service supports setting the crypt salt parameters - see https://ltb-project.org/documentation/self-service-password/1.3/config_ppolicy

PR #72, opened 2019-01-04
WIP: Kopano Konnect und Rest API

fbartels:


konnect/wrapper.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+dockerize \
+	-wait file:///kopano/ssl/konnectd-tokens-signing-key.pem \
+	-wait file:///kopano/ssl/konnectd-encryption.key \
+	-timeout 360s \
+	sleep 2
+exec konnectd serve \
```
longsleep, line 9: That seems unclean - why dockerize sleep?

core/start-service.sh
```diff
@@ -72,6 +72,30 @@ ical)
 	unset "${!KCCONF_@}"
 	exec /usr/sbin/kopano-ical -F
 	;;
+grapi)
+	dockerize \
+		-wait tcp://kopano_server:236 \
```
longsleep: There is no need to wait on server.

docker-compose.yml-example
```diff
@@ -3,7 +3,7 @@ version: "3.5"
 services:
   web:
     image: ${docker_repo:?err}/kopano_web
-    container_name: web
+    container_name: ${FQDN}
```
longsleep: Imho this will only cause issues and makes it impossible to resolve the fqdn externally - strongly suggest to keep this web.

konnect/wrapper.sh
fbartels: Sleep could also be at a dedicated line and not be part of the dockerize command. The problem where the sleep helps is that dockerize continues once the file has appeared. So even though the file is still being written. That in some cases lead to konnect failing to start since the certificate wasn't valid/complete.

longsleep: Mhm certificates and keys should be created atomic. Means they write to a temp file and once complete move the full file to the target which avoids half write issues. How are those files created? Needs to be fixed there instead of here.

Makefile
```diff
@@ -137,13 +140,13 @@ publish-kweb: build-kweb
 	docker push $(docker_repo)/kopano_web:latest
 
 test:
-	docker-compose down -v || true
+	docker-compose -f docker-compose.yml-example down -v || true
```
longsleep: Maybe add the name of the compose file as an overridable make variable ?

core/start-service.sh
```diff
@@ -72,6 +72,27 @@ ical)
...
 	exec /usr/sbin/kopano-ical -F
 	;;
+grapi)
+	LC_CTYPE=en_US.UTF-8
+	export socket_path=/var/run/kopano/grapi
+	mkdir $socket_path
+	chown -R kapi:kopano $socket_path
+	# cleaning up env variables
+	unset "${!KCCONF_@}"
+	exec /usr/sbin/kopano-grapi serve
```
longsleep: It would be nice for all exec lines to use no path - makes it much better to read in the process list and the bin sbin paths should be in the system path list anyways.

docker-compose.yml-example
```diff
@@ -215,18 +201,42 @@ services:
...
+      - kopanossl/:/kopano/ssl
+      - kopanosocket/:/run/kopano
+    environment:
+      - SERVICE_TO_START=kapid
+      - TZ=${TZ}
+      - KCCONF_KAPID_LOG_LEVEL=DEBUG
+      - KCCONF_KAPID_OIDC_ISSUER_IDENTIFIER=http://kopano_konnect:8777 # if this works, then it should go in the python script
+    networks:
+      - kopano-net
+      - web-net
```
longsleep, line 238: Maybe consider renaming those to both have the kopano prefix - why are there two networks anyways?

ssl/start.sh
```diff
@@ -1,7 +1,40 @@
 #!/bin/sh
 
-if [ -f /kopano/ssl/ca.pem ]; then
-	exit 0
+if [ ! -f /kopano/ssl/ca.pem ]; then
```
longsleep, line 5: For all the files which get checked in conditions make sure they are created with atomic operations like mv or ln to avoid half writes and incomplete states. Also ensure that the atomic operation is last. Means always write to tmp files or new files and move or link when all successfully written.

konnect/wrapper.sh
fbartels: yes, that is indeed the better way to solve this. implemented

core/start-service.sh
fbartels: good point

docker-compose.yml-example
fbartels: There is even a third one for ldap. its probably a bit on the paranoid side, but the suggested idea was to have only these containers in the same network, that need to see each other.

as for the prefix, this is already done when the network is through compose:
```
23eb696245db        kopano-docker_kopano-net   bridge              local
21a53ce5d83f        kopano-docker_web-net      bridge              local
```

ssl/start.sh
fbartels: good point. i have adapted the script.

docker-compose.yml-example
```diff
@@ -215,18 +201,42 @@ services:
...
+    image: ${docker_repo:?err}/kopano_core:${CORE_VERSION}
+    container_name: kopano_kapi
+    volumes:
+      - kopanossl/:/kopano/ssl
+      - kopanosocket/:/run/kopano
+    environment:
+      - SERVICE_TO_START=kapid
+      - TZ=${TZ}
+      - KCCONF_KAPID_LOG_LEVEL=DEBUG
+      - KCCONF_KAPID_OIDC_ISSUER_IDENTIFIER=http://kopano_konnect:8777 # if this works, then it should go in the python script
```
longsleep: While this works to load the discovery document, it ultimatively still will not work when the iss as configured within konnect is not resolvable by kapid since the discovery document contains links which need to be loaded (jwks for example).

docker-compose.yml-example
longsleep: ok got - seems a bit complicated for minimal gain but ok

ssl/start.sh
```diff
@@ -1,7 +1,46 @@
...
+	echo "Creating CA and server certificates..."
+	easypki create --filename internalca --organizational-unit primary --expire 3650 --ca "Internal Kopano System"
+
+	for s in kserver kdagent kmonitor ksearch kspooler kwebapp; do
+		easypki create --ca-name internalca --organizational-unit $s --expire 3650 $s
+		cp /kopano/easypki/internalca/keys/$s.key /tmp/$s.pem
+		cat /kopano/easypki/internalca/certs/$s.crt >> /tmp/$s.pem
+		openssl x509 -in /kopano/easypki/internalca/certs/$s.crt -pubkey -noout > /tmp/$s-public.pem
+		cp /tmp/$s.pem /kopano/ssl/$s.pem
+		cp /tmp/$s-public.pem /kopano/ssl/clients/$s-public.pem
```
longsleep: cp is not atomic while rm has the chance to be atomic. So also the choice of location matters. Do not use tmp, use a tmp file in the same folder as the destination to write to and after having created the file, move it to the desired name in the same folder. So no /tmp, no cp.

longsleep: This is for all the create stuff also in other files and down in this script. (Since mv is not atomic when crossing file systems and /tmp is most certainly a different file system like tmpfs).

longsleep (changes requested):

docker-compose.yml-example
fbartels: Yes, I have already realized that. This will need to go back to $fqdn.

ssl/start.sh
fbartels: ok, misconception from my side. now files are created in their location with a tmp file extension and moved afterwards.

longsleep (approved):

Issue #79, opened 2019-01-11
More easy integration with existing reverse proxy


cguentherTUChemnitz:
Hi there again. I am trying to integrate this kopano stack into an existing environment. Therefore i use traefik as a ssl terminating reverse proxy.

What i want to achieve:
- minimal modifications of the docker-compose.yml to be able to integrate as smooth as possible configuration updates

Current situation:
- the kweb service (based on caddy) provides a smooth preconfigured proxy for the kopano related services
- new kopano based services and configuration updates are reflected in the caddy config file

What i try:
-  I try to proxy the caddy with traefik, so i do not have to bother about kopano stack related proxy updates.

The problem:
- Current configuration enforces caddy to serve the services tls encrypted. Proxying those via traefik, leads to internal server errors. (Maybe this is still possible, but i did not manage to get it running.)
- When i patch the kweb.cfg, deleting the tls statement, the proxying works fine.

My proposal:
- I would love to see the kweb "web" service providing an additional entrypoint without tls on an additional port, which is not mapped to the host. So the non-encrypted access is not public by default, but can be accessed by existing reverse proxy infrastructure.

cguentherTUChemnitz:
Update:

Since the caddy related tls config option is already present to be configured via the env var EMAIL, i choosed to use this directly to optionally disable tls on demand during the setup.sh.

fbartels:
Yes, that seem like a clever approach. I was thinking about putting all configuration into a separate file and then use `import` to provide the same configuration for the tls and a hidden non-tls endpoint. But this way it looks cleaner.

Still want to do some tests (probably over the weekend) before merging, though.

cguentherTUChemnitz:
Ok, thanks for looking into it. If you are testing also with traefik, i can provide a starting point here (maybe later on also for documentation purposes):
```
services:
  web:
    labels:
      traefik.frontend.rule: "Host:${FQDN}"
      traefik.protocol: http
      traefik.port: 8443
      traefik.enable: true
      traefik.frontend.headers.forceSTSHeader: true
      traefik.frontend.headers.STSSeconds: 315360000
      traefik.frontend.headers.STSIncludeSubdomains: true
      traefik.frontend.headers.STSPreload: true
      traefik.frontend.whiteList.useXForwardedFor: true
      traefik.docker.network: "proxy-net"


networks:
  proxy-net:
    external: true
```
[...]

PR #80, opened 2019-01-11
External proxy setup

cguentherTUChemnitz:
handles the problems described in https://github.com/zokradonh/kopano-docker/issues/79

Idea:
Use the tls caddy config via setup.sh to optionally disable tls, which is useful when runned behind ssl-terminating reverse proxy.


fbartels:
Finally had some time to try out your change. I just put Nginx in front. Just to not confuse others when looking at the configuration of your reverse proxy I would recommend to use the port 2015 instead of 8443.

PR #81, opened 2019-01-11
enable proxy mode for self service

cguentherTUChemnitz:
i am discussing the correct creation of urls in the mail send by the self-service here:
https://github.com/tiredofit/docker-self-service-password/issues/11

necessary fixes are collected in this pull-request

### DIFF
--- a/docker-compose.yml-example
+++ b/docker-compose.yml-example
@@ -51,7 +51,7 @@ services:
       - web-net
 
   password-self-service:
-    image: tiredofit/self-service-password:3.0
+    image: tiredofit/self-service-password:3.0.1
     container_name: password-self-service
     domainname: ${FQDN}
     depends_on:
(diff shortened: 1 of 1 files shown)


fbartels:
Hi @cguentherTUChemnitz,

as far as I can see there is still some movement in the docker image. is that correct?

cguentherTUChemnitz:
Hi @fbartels, sry for my absence in the last time. I am really busy right know. So i am not able to fix the problem by myself, as well the maintainer seems also to be too busy at the moment to fix it himself. So i would propose, that we deactivate the mail-based self-service and keep only the website-based reset, as long as the url is not embedded correctly in the mail.

fbartels:
Hi @cguentherTUChemnitz ,

no problem. I've glanced over the env variables of the docker image, but could not directly find a "set this env to disable password reset via email" env.

I'd propose to just keep this as it is for now.

fbartels:
I have added some additional lint checks. Please rebase.

fbartels:
closing this for now

PR #82, opened 2019-01-11
WIP: Kopano Meet

fbartels:


konnect/Dockerfile
```diff
@@ -13,5 +13,5 @@ RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSI
     && rm dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz
 
 RUN mkdir -p /etc/kopano/
-COPY identifier-registration.yaml /etc/kopano
+COPY identifier-registration.yaml konnectd-identifier-scopes.yaml /etc/kopano/
```
longsleep: This naming seems a bit inconsistent.

kwmserver/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV CODE_VERSION="${CODE_VERSION}"
+
+USER root
+
+ENV DOCKERIZE_VERSION v0.6.1
+RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
+    && tar -C /usr/local/bin -xzvf dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
+    && rm dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz
+
+COPY wrapper.sh /usr/local/bin
```
longsleep, line 6: Cant this use https://hub.docker.com/r/kopano/kwmserverd ?

web/kweb.cfg
```diff
@@ -97,6 +102,44 @@
...
+    }
+    rewrite /api/v1/websocket {
+        to /upstreams/kwmserver/{path}
+    }
+    rewrite /api/v1/rtm.connect {
+        to /upstreams/kwmserver/{path}
+    }
+    rewrite /api/v1/rtm.turn {
+        to /upstreams/kwmserver/{path}
+    }
```
longsleep: You do not need the kwmserver api/v1/ endpoints for Meet - so best get rid of them. Meet only uses api/kwm/v2

konnect/Dockerfile
fbartels: good point changed

kwmserver/Dockerfile
fbartels: I thought I needed more changes, but by now have also removed dockerize. 
I'd like to keep the wrapper.sh script to easily run kwmserverd in insecure mode for demo environments, though. (which I do not really want to do with the `command` value in the compose file.

web/kweb.cfg
fbartels: cleaned from config

kwmserver/wrapper.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e
+
+if [ "$INSECURE" = "yes" ]; then
+	exec kwmserverd serve \
+		--insecure \
+		--iss="https://$oidc_issuer_identifier"
+else
+	exec kwmserverd serve \
+		--iss="https://$oidc_issuer_identifier"
+fi
```
longsleep, line 11: This if  / else is a little complicated and duplicates code. See the binscripts in scripts directory of the source tree for konnect, kwmserver or kweb on how to extend args based on environment variables without duplicating lines.

meet/start-service.sh
```diff
@@ -0,0 +1,25 @@
...
+ADDITIONAL_KOPANO_PACKAGES=${ADDITIONAL_KOPANO_PACKAGES:-""}
+
+set -eu # unset variables are errors & non-zero return values exit the whole script
+
+if [ ! -e /kopano/$SERVICE_TO_START.py ]; then
+	echo "Invalid service specified: $SERVICE_TO_START" | ts
+	exit 1
+fi
+
+echo "Configure core service '$SERVICE_TO_START'" | ts
```
longsleep: What does "core service" mean?

meet/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    echo "meet-${KOPANO_MEET_VERSION}" > /kopano/buildversion; \
+    # install apt key if supported kopano
+    if [ ${RELEASE_KEY_DOWNLOAD} -eq 1 ]; then \
+        curl -s -S -o - "${KOPANO_MEET_REPOSITORY_URL}/Release.key" | apt-key add -; \
+    fi; \
+    # install
+    apt-get update && \
+    set -x && \
+    apt-get install --no-install-recommends -y \
+        kopano-kwebd \
```
longsleep, line 31: This is a nice idea - we probably should add a dedicated mode to kweb to serve an individual (single) pwa.

kwmserver/wrapper.sh
fbartels: updated

meet/start-service.sh
fbartels: that line was a copy and paste from the "core" docker image. I was a bit torn between just adding the meet relevant stuff on top, or making a "new" image without all the core components inside.

meet/Dockerfile
fbartels: yes, here I also was a bit indifferent. Should it be one docker container/image per pwa or just one including all (future) pwas. I think this can definitely change once there are more apps.

kwmserver/wrapper.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+if [ -n "$oidc_issuer_identifier" ]; then
+	set -- "$@" --iss="https://$oidc_issuer_identifier"
```
longsleep: This looks wrong. A proper OIDC issuer identifer includes "https://" since the protocol it is part of the identifier.

longsleep (changes requested):

kwmserver/wrapper.sh
fbartels: indeed. Before I always prefixed https to the fqdn in the compose file. I have realised this for this case now as well.

kwmserver/Dockerfile
fbartels: On second though now that I changed the variable assignment in wrapper.sh I think we could drop this if this variable assignment would be done in https://hub.docker.com/r/kopano/kwmserverd directly. should I adapt the entrypoint script or do you want to?

longsleep: Go ahead :+1:

PR #94, opened 2019-02-27
Unignore compose

fbartels:
Since the docker-compose.yml is still changing a lot this can lead to some confusion to some users. By adding docker-compose.yml to the versioning these confusions are probably dealt with best.

~~ps: still needs to be rebased to get rid of other commits~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ KOPANO_ZPUSH_REPOSITORY_URL := http://repo.z-hub.io/z-push:/final/Debian_9.0/
 RELEASE_KEY_DOWNLOAD := 0
 DOWNLOAD_COMMUNITY_PACKAGES := 1
 
+COMPOSE_FILE := docker-compose.yml
 -include .env
 export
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -21,7 +21,6 @@ KOPANO_ZPUSH_REPOSITORY_URL := http://repo.z-hub.io/z-push:/final/Debian_9.0/
 RELEASE_KEY_DOWNLOAD := 0
 DOWNLOAD_COMMUNITY_PACKAGES := 1
 
-COMPOSE_FILE := docker-compose.yml-example
```
longsleep, line 24: I would keep the COMPOSE_FILE variable, so that the Mikefile can be used with other compose files.

README.md
```diff
@@ -22,6 +22,10 @@ This repository contains an easy to replicate recipe to spin up a [Kopano](https
 
 The `docker-compose.yml` file by default pulls Docker containers from https://hub.docker.com/r/zokradonh/kopano_core/ and https://hub.docker.com/r/zokradonh/kopano_webapp/. These images are based on the [Kopano nightly builds](https://download.kopano.io/community/) and will contain the latest version available from the time the image was built.
 
+## Updating
+
+Previously the `docker-compose.yml` file was not part of the git repository, which made it harder for users to pick and apply changes in the upstreamed `docker-compose.yml-example` file. This has meanwhile been changed and a `git pull` will now pull changes directly into `docker-compose.yml`. If you need to configure additional env variables, this can now be done in the additional env files (more details further below, for example for `kopano_server` this file is called `kopano_server.env`). If you only want to run a subset of containers it is recommended to create a copy of `docker-compose.yml` and specify your copy when running. e.g. like ´docker-compose -f my-setup.yml up -d´.
```
longsleep: Why is is relevant what previously was? Just explain how it is in this revision :)

longsleep (changes requested):

Issue #95, opened 2019-02-27
Add reverse proxy configuration for /caldav 


fbartels:
This way a secure connection to kopano_ical can be made quite trivial to setup.

PR #96, opened 2019-03-01
Enables caldav via reverse proxy

mhoffmann75:
Enables caldav via reverse proxy on /caldav/ - see https://github.com/zokradonh/kopano-docker/issues/95

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -363,10 +363,12 @@ services:
       - SERVICE_TO_START=ical
       - TZ=${TZ}
       - KCCONF_ICAL_SERVER_SOCKET=http://kopano_server:236/
+      - KCCONF_ICAL_ICAL_LISTEN=*:80
     env_file:
       - kopano_ical.env
     networks:
       - kopano-net
+			- web-net
 
   kopano_monitor:
     image: ${docker_repo:?err}/kopano_core:${CORE_VERSION}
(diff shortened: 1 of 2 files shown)


fbartels:
Ah nice, I must say I secretly hoped that you would pick this up, since you mentioned problems getting ical running with https.

Small comments inline.

Thanks for your pr!

docker-compose.yml
```diff
@@ -363,10 +363,12 @@ services:
       - SERVICE_TO_START=ical
       - TZ=${TZ}
       - KCCONF_ICAL_SERVER_SOCKET=http://kopano_server:236/
+      - KCCONF_ICAL_ICAL_LISTEN=*:80
```
fbartels: this is actually not necessary, since the service already listens on port 8080

web/kweb.cfg
```diff
@@ -152,6 +152,14 @@
         try_duration 30s
     }
 
+    proxy /caldav/ kopano_ical:80 {
```
fbartels: just change port here to 8080

web/kweb.cfg
```diff
@@ -152,6 +152,14 @@
...
+        fail_timeout 10s
+        try_duration 30s
+        transparent
+        keepalive 100
+    }
+    folderish /caldav
```
fbartels, line 160: not quite sure if this is required, but it surely cannot hurt much.

fbartels (changes requested):

mhoffmann75 (commented):
changed as per your suggestions

fbartels (approved):

fbartels:
Thanks again

PR #98, opened 2019-03-05
allow local overrides to docker-compose

mhoffmann75:
Exclude docker-compose.override.yml file from git updates to allow for local customizations of projects docker-compose.yml file


fbartels:
Ah, I wasn't aware of this feature until now.

Thanks for your pr.

PR #104, opened 2019-03-10
Changed 'sed -r' to 'sed -E' in setup.sh for better compatibility

r00tc0d3:
With this change it also works on a Mac. Also tested it on Arch Linux


fbartels:
Thanks for your pr!

PR #106, opened 2019-03-11
Make remaining mail ports (SMTPS/MSA/IMAP) configurable.

mithomas:
While testing this there were some clashes with existing services - this PR is supposed to make all the mail-related ports configurable to avoid such issues.


docker-compose.yml
```diff
@@ -110,8 +110,8 @@ services:
       - ldap
     ports:
       - "${SMTPPORT:-25}:25"
-      - "465:465"
-      - "587:587"
+      - "${SMTPSPORT:-465}:465"
+      - "${MSAPORT:-587}:587"
```
fbartels, line 114: What does MSA stand for? Mail Submission A...?

Maybe a better name would just be $SUBMISSIONPORT ?

fbartels (changes requested):
Hi @mithomas,

thanks for your pr. Generally it looks good to me. Have just a small comprehension question.

docker-compose.yml
mithomas: It stands for _Mail Submission Agent_ (as per https://stackoverflow.com/questions/15796530/what-is-the-difference-between-ports-465-and-587). In [IANA's service name registry](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=587), this is indeed called _submission_ - I'll adjust that.

fbartels: Ah, never heard that abbreviation. IANA also just refers to it as submission, but since this seems to be an accepted term on stackoverflow its probably good here as well.

Thanks for the clarification, will mere now.

fbartels (approved):

PR #107, opened 2019-03-11
switch restart always to unless-stopped

cguentherTUChemnitz:
I try to integrate the backup mechanisms of https://github.com/blacklabelops/volumerize, handling named volumes and automatic service stop and restart for consistent backups. Nevertheless this needs a clean stop of a service, which is only able to be achieved, if the docker restarts the containers only if they are not explicitly stopped.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   web:
     image: ${docker_repo:-zokradonh}/kopano_web
     container_name: web
-    restart: always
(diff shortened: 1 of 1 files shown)


fbartels:
Hi @cguentherTUChemnitz,

While `unless-stopped` is probably the better option (one could argue that most of the other containers could probably benefit from it as well) I strongly suggest to not stop the containers for backup. Stopping means that you loose accumulated caches.

PR #109, opened 2019-03-11
Make real IP logging work in proxy scenario

mhoffmann75:
USE_CUSTOM_REMOTE_IP_HEADER should either be false or contain the name of the header to be used.


fbartels:
Thanks for your pr.

PR #111, opened 2019-03-13
Adding GABSync functionality

mhoffmann75:
Providing extra container with z-push-GABSync run every 15 minutes to keep GAB for Kopano Outlook Extension in sync

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -273,6 +273,21 @@ services:
       - web-net
       - kopano-net
 
+  kopano_gabsync:
+    image: ${docker_repo:-zokradonh}/kopano_gabsync:${ZPUSH_VERSION:-latest}
+    hostname: kopano_gabsync
+    container_name: kopano_gabsync
+    volumes:
+      - kopanossl/:/kopano/ssl
+      - kopanosocket/:/run/kopano
+    environment:
+      - TZ=${TZ}
+    env_file:
+      - kopano_gabsync.env
+    networks:
+      - web-net
+      - kopano-net
+
   kopano_grapi:
     image: ${docker_repo:-zokradonh}/kopano_core:${CORE_VERSION:-latest}
     container_name: kopano_grapi
--- a/gabsync/Dockerfile
+++ b/gabsync/Dockerfile
@@ -1,0 +1,72 @@
+ARG docker_repo=zokradonh
+FROM ${docker_repo}/kopano_base
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+ARG ADDITIONAL_KOPANO_PACKAGES=""
+ENV ADDITIONAL_KOPANO_PACKAGES=$ADDITIONAL_KOPANO_PACKAGES
+ARG DOWNLOAD_COMMUNITY_PACKAGES=1
+ENV DOWNLOAD_COMMUNITY_PACKAGES=$DOWNLOAD_COMMUNITY_PACKAGES
+ARG KOPANO_CORE_REPOSITORY_URL="file:/kopano/repo/core"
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -273,6 +273,21 @@ services:
...
+    container_name: kopano_gabsync
+    volumes:
+      - kopanossl/:/kopano/ssl
+      - kopanosocket/:/run/kopano
+    environment:
+      - TZ=${TZ}
+    env_file:
+      - kopano_gabsync.env
+    networks:
+      - web-net
```
fbartels, line 289: this container would not need to be exposed to the proxy instance

gabsync/Dockerfile
```diff
@@ -0,0 +1,72 @@
+ARG docker_repo=zokradonh
+FROM ${docker_repo}/kopano_base
```
fbartels, line 2: instead of duplicating much of the code from z-push container this should use the z-push container as a base.

fbartels: maybe this container could be called kopano_zpush_utils (similar to the kopano_utils)?

gabsync/start.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+echo "Starting Gabsync"
+# cleaning up env variables
+unset "${!KCCONF_@}"
+while true; do
+  set +e
+  /usr/share/z-push/tools/gab-sync/gab-sync.php -a sync
+  set -e
+  echo "Automatically running GABSync in 15 minutes - restart container to rerun it immediately"
```
fbartels, line 35: there is really no need to run gabsync that often. as explained in https://wiki.z-hub.io/display/ZP/Configuring+GAB-Sync+for+Kopano+OL+Extension#ConfiguringGAB-SyncforKopanoOLExtension-Automatingexecutionofz-push-gabsync it should be sufficient to run it once per day. if you want your user changes be taken up faster then this should be integrated into the userscripts of kopano_core.

setup.sh
```diff
@@ -65,7 +65,7 @@ docker_tag_search () {
 }
 
 echo "Creating individual env files for containers (if they do not exist already)"
-for dockerenv in ldap password-self-service mail db kopano_ssl kopano_server kopano_webapp kopano_zpush kopano_grapi kopano_kapi kopano_dagent kopano_spooler kopano_gateway kopano_ical kopano_monitor kopano_search kopano_konnect kopano_kwmserver kopano_meet; do
+for dockerenv in ldap password-self-service mail db kopano_ssl kopano_server kopano_webapp kopano_zpush kopano_gabsync kopano_grapi kopano_kapi kopano_dagent kopano_spooler kopano_gateway kopano_ical kopano_monitor kopano_search kopano_konnect kopano_kwmserver kopano_meet; do
```
fbartels, line 68: changes to the makefile would also be required

fbartels (changes requested):
Hi @mhoffmann75,

thanks for the pr. I do agree that it would be nice to have a way the execute the gabsync utility, but I do not really like the current approach. Comments inline.

PS: you should rebase against upstream master to get rid of the surplus merge commit.

fbartels:
I have added some additional lint checks. Please rebase.

docker-compose.yml
mhoffmann75: Yep you're right.

gabsync/Dockerfile
mhoffmann75: I would love to name it kopano_gabsync because of the container = micro service idea. If we need some other sync tool later on it should be deployed in a separate container aswell.

mhoffmann75:
making new PR

PR #112, opened 2019-03-16
make webapp configurable from env

fbartels:
inspired by https://github.com/mlan/docker-kopano/blob/master/assets/entrypoint.sh#L80-L92 and https://github.com/zokradonh/kopano-docker/blob/a7c213434766f5433148ea6b50e0894094b92e12/meet/start-service.sh#L22-L27

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,11 +206,14 @@ check-scripts:
 	git ls-files --exclude='Dockerfile*' --ignored | xargs --max-lines=1 hadolint
 
 test:
-	docker-compose -f $(COMPOSE_FILE) down -v || true
+	docker-compose -f $(COMPOSE_FILE) down -v --remove-orphans || true
 	make build-all
 	docker-compose -f $(COMPOSE_FILE) build
 	docker-compose -f $(COMPOSE_FILE) up -d
 	docker-compose -f $(COMPOSE_FILE) ps
+
+test-update-env:
+	docker-compose -f $(COMPOSE_FILE) up -d
 
 test-quick:
 	docker-compose -f $(COMPOSE_FILE) stop || true
--- a/README.md
+++ b/README.md
@@ -32,11 +32,52 @@ Yes, indeed. You could for example use this to easily try out newer Kopano WebAp
 
 And last but not least this project also offers a `zokradonh/kopano_utils` image to easily run tools such as `kopano-backup`, `kopano-migration-pst`, `kopano-migration-imap` and all the other utilities that are bundles with Kopano. See [below](#some-more-commands-for-those-unfamilar-with-docker-compose) to see how to run `zokradonh/kopano_utils`.
 
-### Need to adjust any values after the initial run of `setup.sh`?
+### Additional configuration / Need to adjust any values after the initial run of `setup.sh`?
 
 If you want to modify some of the values from the `setup.sh` run you can simply edit `.env` in your favourite editor. Repeated runs of `setup.sh` will neither modify `docker-compose.yml` nor `.env`. In the ´.env´ file you will also find some given defaults like ldap query filters and the local ports for the reverse proxy.
 
 Additionally a dedicated env file is created for each container (at least where that would make sense). The env file has the container name as part of the file name. For example for the `kopano_server` container the filename is `kopano_server.env`. These additional env files are auto created when running `setup.sh`
+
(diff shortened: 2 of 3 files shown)


fbartels:
@cguentherTUChemnitz can you also give this a try?

cguentherTUChemnitz:
Thanks for your affort. I try to test it this Weekend.

cguentherTUChemnitz:
@fbartels in which way do i have to provide the environment variables? Currently i tried to the `kopano_webapp.env`:

```
KCCONF_WEBAPPPLUGIN_PLUGIN_MDM_USER_DEFAULT_ENABLE_MDM=localhost
KCCONF_WEBAPPPLUGIN_PLUGIN_MDM_SERVER_SSL=true
KCCONF_WEBAPPPLUGIN_PLUGIN_MDM_USER_DEFAULT_ENABLE_MDM=true
```

But those values were not reflected in the container internal config.php `/usr/share/kopano-webapp/plugins/mdm/config.php`. This file still does contain its default values.

Did i misinterpret how i have to compose the env var names?

fbartels:
Ah, I should have mentioned this. 

For webapp itself the envs are prefixed with KCCONF_WEBAPP_ then follows the config option, as it is in config.php the value is the value you want to set in the file.

For the plugins its a bit different, the general prefix is KCCONF_WEBAPPPLUGIN_ , then follows the file you want to configure, e.g. MEET_ and then the config option as it is in the config file.

So to enable the mdm plugin by default it should be `KCCONF_WEBAPPPLUGIN_MDM_PLUGIN_MDM_USER_DEFAULT_ENABLE_MDM=true`

cguentherTUChemnitz:
I do appreciate a lot your added documentation. Especially your ascii art is very nice ;)

Nevertheless, i am currently not able to produce the entry in the desiged php config file. I added the config from your documented example to the kopano_webapp.env:
```
KCCONF_WEBAPPPLUGIN_MDM_PLUGIN_MDM_USER_DEFAULT_ENABLE_MDM=true
```

When i stop, rm and up this service, than i still have the default config in /etc/kopano/webapp/config-mdm.php:
```
<?php
define('PLUGIN_MDM_USER_DEFAULT_ENABLE_MDM', false);
define('PLUGIN_MDM_SERVER', 'localhost');
define('PLUGIN_MDM_SERVER_SSL', false);
?>
```

fbartels:
Thanks for taking it for a spin, I found my issue. The identifier was still prefixed in the config option sed was using for the plugins. (also the webapp plugins envs were also found when updating config.php).

I have updated the pr with this fix.

cguentherTUChemnitz:
Do i test correctly? I do the following stuff:
- I am using the latest fbartels/php-config branch.
- I have an running kopano-stack and for faster testing i kill and remove only the kopano_webapp.
- i add `KCCONF_WEBAPPPLUGIN_MDM_PLUGIN_MDM_USER_DEFAULT_ENABLE_MDM=true` to kopano_webapp.env
- I run docker-compose up -d
- After the webapp has started the apache server i check the config-mdm.php settings with `docker exec -it kopano_webapp cat /etc/kopano/webapp/config-mdm.php`

Currently the result is still incorrect with:
```
<?php
define('PLUGIN_MDM_USER_DEFAULT_ENABLE_MDM', false);
define('PLUGIN_MDM_SERVER', 'localhost');
define('PLUGIN_MDM_SERVER_SSL', false);
?>
```

fbartels:
No, you are missing a small, but important step.

- checkout the fbartels/php-config branch
  - while being on master you could run the following two commands:
    ```
    git checkout -b fbartels-php-config master
    git pull https://github.com/fbartels/kopano-docker.git php-config
    ```
- run `make build-webapp` to rebuild webapp with the changes from this branch
- add your changes to `kopano_webapp.env`
  - for testing I had in mine the following:
    ```
    #KCCONF_SERVERHOSTNAME=my.prod.setup.com
    KCCONF_WEBAPP_CLIENT_TIMEOUT=3600
    KCCONF_WEBAPPPLUGIN_MDM_PLUGIN_MDM_USER_DEFAULT_ENABLE_MDM=true
    ```
- run `docker-compose up -d` to recreate the webapp container based on the new env (you could also run `make test-update-env` I've added this in this branch to easily retest webapp)
- check `docker-compose logs -f kopano_webapp`
- check `docker-compose exec kopano_webapp cat /etc/kopano/webapp/config-mdm.php`

cguentherTUChemnitz:
Oh thanks for the hint. So i run my tests against my locally cached images not containing your changes 'D. making this image, now provides me the possibility to override the settings correctly. Thanks, very nice.

fbartels:
Ok, then "go" for merge.

PR #115, opened 2019-03-18
handle true/false differently in webapp config

fbartels:


cguentherTUChemnitz:
Your assumption of the quotes is right. The configuration value is now inserted correctly without the quotes:
```
define('INSECURE_COOKIES', true);
```
Consequentially the error behind my ssl-terminating reverse-proxy is gone.

PR #116, opened 2019-03-19
Add Gabsync

mhoffmann75:
Provides extra container to run Gabsync every 24hours.

### DIFF
--- a/webapp/start.sh
+++ b/webapp/start.sh
@@ -15,7 +15,15 @@ php_cfg_gen() {
 	if [ -e "$cfg_file" ]; then
 		if grep -q "$cfg_setting" "$cfg_file"; then
 			echo "Setting $cfg_setting = $cfg_value in $cfg_file"
-			sed -ri "s#(\s*define).+${cfg_setting}.+#\define(\x27${cfg_setting}\x27, \x27${cfg_value}\x27\);#g" "$cfg_file"
+			case $cfg_value in
+			true|TRUE|false|FALSE)
+				echo boolean value
+				sed -ri "s#(\s*define).+${cfg_setting}.+#\tdefine(\x27${cfg_setting}\x27, ${cfg_value}\);#g" "$cfg_file"
+				;;
+			*)
+				sed -ri "s#(\s*define).+${cfg_setting}.+#\tdefine(\x27${cfg_setting}\x27, \x27${cfg_value}\x27\);#g" "$cfg_file"
+				;;
+			esac
 		else
 			echo "Error: Config option $cfg_setting not found in $cfg_file"
 			cat "$cfg_file"


fbartels:
Hi @mhoffmann75,

thanks for your pr. I have looked through it but will need to think about it for a while. Like already said in #111 I do see that there is a need for this, but I am still unsure about the approach.

Instead of having an own image for gabsync I think the changes could even do directly into the z-push image and onc could enhance the start.sh script with something like https://github.com/zokradonh/kopano-docker/blob/master/core/start-service.sh#L32-L36

This way you can just call `docker-compose exec kopano_zpush /usr/share/z-push/tools/gab-sync/gab-sync.php -a sync` (I think that should work then, did not try it).

Also still not sure (even though I do it myself in https://github.com/zokradonh/kopano-docker/blob/master/core/services/kopano-users.sh) if the trigger to do something on a regular base should come from inside of the container running the task.

Maybe a general task scheduler (for example based on https://github.com/aptible/supercronic) would be interesting?

I am open for ideas.

mhoffmann75:
I think this is a matter of philosophy. The docker philosophy from my understanding is about micro services. So we should *not* create containers running multiple services unless forced to - e.g. mailserver is such an example. We should create a container for each service / task. For example you don't need gab sync because you don't plan to use KOE or later on its not needed anymore, just stop and remove the container. This is much easier than fiddling with the service from cron inside a container. Besides this "cron" container would need to have all components installed inside that are needed for various tasks making it heavy and you would have components inside you probably don't need at all.
So my vote is for micro service architecture making things granular and "simple". That is why I insisted on naming it kopano_gabsync instead of your suggestion kopano_zpush_utils - see https://github.com/zokradonh/kopano-docker/pull/111#discussion_r266970092

mhoffmann75:
Just another thought: It seems to me that one thing that makes this container unnecessary heavy is the dependencies between the deb packets. From my understanding we need **z-push-kopano-gabsync** which required PHP Mapi which we can get from the packet **z-push-backend-kopano**. However the latter pulls in php (OK) and full apache server. So I'm not sure if this is really needed!? 
That's the reason why my first approach was to built this container from kopano_base instead of kopano_zpush not realizing that it pulls in full z-push stack :-)
But you're right its easier to create and maintain it from zpush as template. Maybe with some litte additional overhead inside the container - maybe not. I have not yet looked over all the dependencies.

fbartels:
> It seems to me that one thing that makes this container unnecessary heavy is the dependencies between the deb packets

yes, they indeed have not been designed with the thinking to be used completely separate (if you want z-push-admin then the usual way would be to have z-push on the same system as well). The apache part wonders me a bit though, as there is a dedicated config package for the webserver stuff.

> The docker philosophy from my understanding is about micro services. So we should not create containers running multiple services unless forced

totally agree.

> Besides this "cron" container would need to have all components installed inside that are needed for various tasks making it heavy and you would have components inside you probably don't need at all.

Maybe I was not clear enough in expressing my vision.

For the cron container I was really just thinking about a job scheduler in a separate container that then triggers commands in either running containers or spins up a new container to run a specific task. (but for this scheduler to be able to interact with Docker it would need to have access to the docker socket, also not really optimal)

Just like you can currently already run `docker run --rm -it -v /var/run/kopano/:/var/run/kopano -v $(pwd):/kopano/path zokradonh/kopano_utils kopano-backup` to create a backup. while backup runs there is a container and afterwards it gets destroyed. Or to stay within the reach of docker-compose `docker-compose exec kopano_server kopano-cli --list-users` to list users.

Similarly to how the kopano_core docker image is started multiple times all the needed packages for the gabsync could be installed and configured in the normal z-push image and then you define another container in docker-compose.yml that then just uses `/usr/share/z-push/tools/gab-sync/gab-sync.php -a sync` as its command. While the job runs the container is there and then it stops until the job scheduler starts it up again.

mhoffmann75:
Ok i will see if a dedicated container with supercronic will work for this task.

PR #117, opened 2019-03-20
Remove redundant explicit name for ldap-net.

mithomas:
Since the custom name for the network ldap-net is identical to the default, this line is redundant.


fbartels:
Thanks for your pr!

PR #118, opened 2019-03-20
Prefix container names with COMPOSE_PROJECT_NAME.

mithomas:
This way it is easily possible to use several instances of this service stack on one host via different .env-files (e.g. for staging/production).
Volumes and networks are separated in this manner by docker-compose anyway, now the containers are too.


docker-compose.yml
```diff
@@ -194,7 +194,7 @@ services:
   kopano_server:
     image: ${docker_repo:-zokradonh}/kopano_core:${CORE_VERSION:-latest}
     hostname: kopano_server
-    container_name: kopano_server
+    container_name: ${COMPOSE_PROJECT_NAME}-kopano_server
     depends_on:
       - db
```
fbartels, line 199: would't this change also require changes to the depends_on values_

docker-compose.yml
```diff
@@ -162,7 +162,7 @@ services:
   db:
     image: mariadb:10.3.10-bionic
     restart: unless-stopped
-    container_name: kopano_db
+    container_name: ${COMPOSE_PROJECT_NAME}-kopano_db
```
fbartels: docker networks use ${COMPOSE_PROJECT_NAME}_network-name as a format. if this is done for consistency, then this should imho also use an underscore here as well.

fbartels: On a related note, make it rather ${COMPOSE_PROJECT_NAME}_db than the current one. else we would currently end up with kopano-kopano_db as the container name in this example.

docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: "3.5"
 services:
   web:
     image: ${docker_repo:-zokradonh}/kopano_web
-    container_name: web
+    container_name: ${COMPOSE_PROJECT_NAME}-web
     restart: unless-stopped
     ports:
       - "2015:2015"
```
fbartels: if the goal is to run this multiple times on the same box, this hardcoded port should probably go into a variable as well.

fbartels (changes requested):
Hi @mithomas,

thanks for your pr. I've made some review comments.

docker-compose.yml
fbartels: to answer myself, no it wouldn't since this value references the name of the container in compose, and not the value in container_name

cguentherTUChemnitz:
I think i should explain some docker-compose naming mechanics here. I hope i start with it a fruitful discussion in which situation we should use names and when not. Docker-compose does some automatic name mangling and demangling which are very helpful for specific use-cases.

Default behavior:
When not specifying names for containers, networks and volumes, docker-compose will generate such based on the COMPOSE_PROJECT_NAME. This project name defaults to the folder-name, where the docker-compose.yml lives in. This project name is prepended with an underscore, than it follows the name of the service (network, volume). In the case of handling services, docker-compose does also append an underscore and the instance number of the service. So you do not have to specify the full name in your compose file, docker-compose does the demangling for you and you can target directly the "local" names. If you want to see the real names exposed by the docker daemon you should look at the docker commands listing the service, networks, volumes.

There are some interesting use-cases you can have with this mechanism. Assume you have a service stack, where all volumes, networks and services do have automatic names, generated by docker-compose, than you are able to spin up the same stack in a second environment on the same docker-host. Therefore you only have to copy your docker-compose.yml into a folder with a different name and all services, volumes and networks are generated with a different prependet project name. This can be used, e.g. to  spin up a testing environment on the same machine where the production one is hosted.

So this leads me to the next point. Specifying the names explicity in the compose file deactivates this automatic name generation and you have to ensure manually to not use the same names multiple times. So for example this changeset does have a lot of impact in such a situation:
https://github.com/zokradonh/kopano-docker/pull/117

If you run `docker network ls`, you see that the name of the network was expcitly set to ldap-net without the usage of the automatic project name prepending. After this change you should notice that the networks real name exposed by `docker network ls` "${COMPOSE_PROJECT_NAME}_ldap-net". Inside the kopano service stack the ldap-net does not have any difference, because docker-compose does demangle the name "${COMPOSE_PROJECT_NAME}_ldap-net" to "ldap-net" automatically. 

So in conclusion i would vote for a solution, where all names are not defined in the docker-compose.yml, except for them you can assume, that they are referenced also by other services, like the ldap-net.

docker-compose.yml
mithomas: True, did not occur to me since I ran my stack without the web-service. Why is this port exposed anyway?

docker-compose.yml
mithomas: Also changed for ssl, server, webapp, zpush, grapi, kapi, kdav, dagent, spooler, gateway, ical, monitor, search, konnect, playground, kwmserver and meet.

docker-compose.yml
fbartels: A fallback (it's the default port for caddy) it's also useful and less confusing when running web without ssl)

mithomas:
> So in conclusion i would vote for a solution, where all names are not defined in the docker-compose.yml, 

I do not know the reason why the containers got specific names in the first place and thus did not try to get rid of them altogether. I do expect there _is_ a reason, though,  since I got no review comment pointing out that the names are obsolete anyway :-). Personally, I do not have a preference on that matter.

> except for them you can assume, that they are referenced also by other services, like the ldap-net.

I would suggest not to assume such a use case by default - I can think of many setups where the used LDAP server lies outside of this service stack. Adding a specific network name in an override is quite easy, should this actually be wanted - removing it is for more inelegant.

fbartels:
>  I do expect there is a reason, though, since I got no review comment pointing out that the names are obsolete anyway

I'm afraid that this is simply a case of "it has always been like that". Not quite sure when it was introduced (I did not bother to check the git log) but since this created stable container names I continued doing so.

There are a few places in the code that use these hard coded names, but I am quite sure these could be made dynamic as well.

For example https://github.com/zokradonh/kopano-docker/blob/master/core/start-service.sh#L54

fbartels:
Reading https://docs.docker.com/compose/compose-file/ there may be some benefit to have a hard coded container name for some of the resources.

> container_name
Specify a custom container name, rather than a generated default name.
>
> container_name: my-web-container
Because Docker container names must be unique, you cannot scale a service beyond 1 container if you have specified a custom name. Attempting to do so results in an error.

Some resources, like for example kopano_server, or the database can not be scaled by just staring additional instances. To seamlessly scale webapp one would need to share php and webapp sessions between instances (I thought I created a ticket for that already, but cannot find it at the moment). to scale z-push one should put states into a mysql database (concurrent file access to this can lead to locks). similar things probably apply to the kdav container. other containers like ical, gateway, web, meet, konnect, .. will probably have no issues with just staring additional instances.

mithomas:
Is this an adjustment for this PR or should this be though over and changed in another one?

fbartels (approved):

fbartels:
I think its fine to pick up this discussion in its own topic. With your changes all containers still start for me, so its good to merge.

PR #119, opened 2019-03-20
Add dynamic dependency-wait for server and spooler service.

mithomas:
Instead of using hard-coded dependency-waits on startup for database and MTA for the server and spooler, respectively, the actually configured values are now used.

For the servers, this also takes the use of a unix file socket over a network socket into account.

### DIFF
--- a/core/start-service.sh
+++ b/core/start-service.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 ADDITIONAL_KOPANO_PACKAGES=${ADDITIONAL_KOPANO_PACKAGES:-""}
+KCCONF_SERVER_MYSQL_SOCKET=${KCCONF_SERVER_MYSQL_SOCKET:-""}
 
 set -eu # unset variables are errors & non-zero return values exit the whole script
 
@@ -39,7 +40,11 @@ fi
 case "$SERVICE_TO_START" in
 server)
 	# determine db connection mode (unix vs. network socket)
-	if [ -n "$KCCONF_SERVER_MYSQL_SOCKET" ]; then DB_CONN="file://$KCCONF_SERVER_MYSQL_SOCKET"; else DB_CONN="tcp://$KCCONF_SERVER_MYSQL_HOST:$KCCONF_SERVER_MYSQL_PORT"; fi
+	if [ -n "$KCCONF_SERVER_MYSQL_SOCKET" ]; then
+		DB_CONN="file://$KCCONF_SERVER_MYSQL_SOCKET"
+	else
+		DB_CONN="tcp://$KCCONF_SERVER_MYSQL_HOST:$KCCONF_SERVER_MYSQL_PORT"
+	fi
 	/kopano/services/kopano-public-store.sh &
 	/kopano/services/kopano-users.sh &
 	dockerize \
--- a/core/start-service.sh
+++ b/core/start-service.sh
@@ -38,12 +38,14 @@ fi
 # start regular service
 case "$SERVICE_TO_START" in
 server)
+	# determine db connection mode (unix vs. network socket)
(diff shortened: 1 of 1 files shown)


fbartels:
Hi @mithomas,

thanks for your pr. Can you fix the current test failure caused by your change? Afterwards it's good to be merged.

```
$ make check-scripts
grep -rIl '^#![[:blank:]]*/bin/\(bash\|sh\|zsh\)' \
	--exclude-dir=.git --exclude=*.sw? \
	| xargs shellcheck
In core/start-service.sh line 120:
		-wait tcp://$KCCONF_SPOOLER_SMTP_SERVER:25 \
                            ^-- SC2086: Double quote to prevent globbing and word splitting.
make: *** [check-scripts] Error 123
```

core/start-service.sh
```diff
@@ -115,7 +117,7 @@ search)
 spooler)
 	dockerize \
 		-wait file://var/run/kopano/server.sock \
-		-wait tcp://mail:25 \
+		-wait tcp://$KCCONF_SPOOLER_SMTP_SERVER:25 \
```
fbartels: Actually one more thing. From the looks of this, this either needs to have a default value or needs to be added to docker-compose.yml (with a default value matching the old case) , else startup will fail for everyone after merge.

fbartels (changes requested):

core/start-service.sh
fbartels: Wait nevermind. this is already in compose.

fbartels (approved):

fbartels:
With this change the core container fails to start:

```
$ docker-compose logs kopano_server
Attaching to kopano_server
kopano_server            | Mar 22 07:52:35 Configure core service 'server'
kopano_server            | Mar 22 07:52:36 Set ownership
kopano_server            | /kopano/start-service.sh: line 42: KCCONF_SERVER_MYSQL_SOCKET: unbound variable
```

I just now also noticed that the if to check KCCONF_SERVER_MYSQL_SOCKET is all in one single line. Lines of code is not of much concern in this place, I would rather recommend to split this over multiple lines. (I would have changed this after merging your pr, but maybe since this still requires work you want to change this as well).

fbartels (changes requested):
```
diff --git a/core/start-service.sh b/core/start-service.sh
index 68275b9..3234491 100755
--- a/core/start-service.sh
+++ b/core/start-service.sh
@@ -1,6 +1,7 @@
 #!/bin/bash

 ADDITIONAL_KOPANO_PACKAGES=${ADDITIONAL_KOPANO_PACKAGES:-""}
+KCCONF_SERVER_MYSQL_SOCKET=${KCCONF_SERVER_MYSQL_SOCKET:-""}

 set -eu # unset variables are errors & non-zero return values exit the whole script

@@ -39,7 +40,11 @@ fi
 case "$SERVICE_TO_START" in
 server)
        # determine db connection mode (unix vs. network socket)
-       if [ -n "$KCCONF_SERVER_MYSQL_SOCKET" ]; then DB_CONN="file://$KCCONF_SERVER_MYSQL_SOCKET"; else DB_CONN="tcp://$KCCONF_SERVER_MYSQL_HOST:$KCCONF_SERVER_MYSQL_PORT"; fi
+       if [ -n "$KCCONF_SERVER_MYSQL_SOCKET" ]; then
+               DB_CONN="file://$KCCONF_SERVER_MYSQL_SOCKET"
+       else
+               DB_CONN="tcp://$KCCONF_SERVER_MYSQL_HOST:$KCCONF_SERVER_MYSQL_PORT"
+       fi
        /kopano/services/kopano-public-store.sh &
        /kopano/services/kopano-users.sh &
        dockerize \
```
would fix the current broken startup

fbartels (approved):

PR #123, opened 2019-03-22
wip: Implement a scheduler to run recurring tasks (such as z-push-gabsync)

### LINKED ISSUE #121: Come up with a way to run z-push-gabsync
as raised in #116. some comments are already in the ticket.

For the moment I am going to assume that @mhoffmann75 is still busy with this.

fbartels:
Code originally from @mhoffmann75 

Open tasks:
[ ] get crontab from env
[ ] build, tag and push the image in makefile
[ ] also run kopano-cli --sync through this

fixed #121

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ export
 # convert lowercase componentname to uppercase
 COMPONENT = $(shell echo $(component) | tr a-z A-Z)
 
-build-all: build-base build-core build-kdav build-konnect build-kwmserver build-ldap-demo build-meet build-playground build-ssl build-utils build-web build-webapp build-zpush
+build-all: build-base build-core build-kdav build-konnect build-kwmserver build-ldap-demo build-meet build-playground build-scheduler build-ssl build-utils build-web build-webapp build-zpush
 
 .PHONY: build
 build: component ?= base
@@ -79,6 +79,9 @@ build-playground:
 build-kdav:
 	component=kdav make build
 
+build-scheduler:
+	component=scheduler make build-simple
+
 build-ssl:
 	component=ssl make build-simple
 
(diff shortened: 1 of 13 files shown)


fbartels:
@mhoffmann75 what do you think of the current state? I think I overengineered it a bit :-D

the gabsync is currently still failing with a permission denied. have not looked into the root cause, yet.

fbartels:
ah of course, the gabsync cfg needs to be changed...

mhoffmann75:
Quite promising. I really like the env approach to dynamically built crontab. Really good work!

It could be just a little confusing that each job is started on container start. From a kopano perspective this may be alright. But think you wanted to add kopano_backup to scheduler. This would be a surprise if it always run on startup?!

Maybe either we should add a statically crontab file that is merged inside the container to get user defined jobs run (maybe through docker-compose.override.yml env) or add a custom format to the Cron spec like e.g.:
@start * * * * docker exec xxx

These would be just executed once but never written to supercronics crontab. 
This way you could also handle public store creation. 

What do you think?

mhoffmann75:
Wow that was quick:-)

mhoffmann75:
Shouldn't start.sh delete/wipe it's cronfile on startup before adding variables? In case container is rerun wouldn't it stack up jobs?

fbartels:
> Shouldn't start.sh delete/wipe it's cronfile on startup before adding variables?

that's a good point. I have only tested with repeated `docker-compose up -d` this completely recreated the container and therefore duplication wasn't an issue. fixed now.

>  add a custom format to the Cron spec

I think its a benefit to test if the configured job will actually succeed, therefore i added this one off execution to the start. but your point about using this for backup is valid. the easier route for the moment was to simply have a second class of cron jobs that will not be executed at startup.

I have added some documentation to the readme that also explains this.

mhoffmann75:
Good solution I think.

PR #131, opened 2019-03-25
replace apache and mod_php with kweb and php-fpm

### LINKED ISSUE #36: performance: use php-fpm instead of mod_php in web containers
- [x] webapp
- [x] z-push
- [ ] kdav

fbartels:
fixes #36

TODO:
- [x] serve webapp through php-fpm
- [ ] make the same change for z-push
- [ ] make the same change for kdav


webapp/extra.cfg
```diff
@@ -0,0 +1,8 @@
+errors stdout
```
longsleep: Should be `stderr` no?

webapp/Dockerfile
```diff
@@ -66,39 +66,31 @@ RUN \
...
+    mkdir -p /run/php && chown www-data:www-data /run/php && \
+    crudini --set /etc/php/7.0/fpm/php.ini PHP upload_max_filesize 500M && \
+    crudini --set /etc/php/7.0/fpm/php.ini PHP post_max_size 500M && \
+    crudini --set /etc/php/7.0/fpm/php.ini PHP max_input_vars 1800 && \
+    crudini --set /etc/php/7.0/fpm/php.ini Session session.save_path /run/sessions
 
 EXPOSE 80/tcp
 
 COPY start.sh /kopano/start.sh
+COPY extra.cfg /etc/kopano/kweb/extra.d/extra.cfg
```
longsleep: This serves only webapp, so consider using caddy mode with a Caddyfile.

longsleep: Or maybe not, since then kweb.cfg is not really used much.

webapp/start.sh
```diff
@@ -76,14 +76,19 @@ for setting in $(compgen -A variable KCCONF_WEBAPPPLUGIN_); do
...
-echo "Starting Apache"
-rm -f /run/apache2/apache2.pid
-set +u
-# shellcheck disable=SC1091
-source /etc/apache2/envvars
 # cleaning up env variables
 unset "${!KCCONF_@}"
-exec /usr/sbin/apache2 -DFOREGROUND
+echo "Starting php-fpm"
+/usr/sbin/php-fpm7.0 -F &
```
longsleep: It would look nicer in the process tree (like ps), if you would call long running processes without their paths.

webapp/extra.cfg
fbartels: +1

webapp/Dockerfile
fbartels: I see we had the same idea. its now a dedicated caddyfile.

webapp/start.sh
fbartels: +1

fbartels:
replaced by #138

PR #132, opened 2019-03-27
Prefix scheduler container name with COMPOSE_PROJECT_NAME.

mithomas:
In line with the other names for running multiple instances of this stack on the same host.


fbartels:
Thanks for your pr. That was the force of habit :-D

I have also fixed the env_file name and merged your pr manually.

PR #166, opened 2019-05-31
wip: add script to reload configuration of running (kopano-server) container

fbartels:
this generally works but has the downside that values from .env and kopano_server.env are only reread when running docker-compose up and not when running docker-compose exec kopano_server reload-service.sh.

command will need to be something like:
docker-compose exec -e configoption=new-value kopano_server reload-service.sh

Not quite happy with it yet. Putting it here to not forget about it (and maybe someone else has some insights).


fbartels:
Idea: put all env variables into file. docker cp file into container. source file in reload script. send hup to service.

zokradonh:
Some thoughts and ideas about it:

I feel better with `docker kill --signal=SIGHUP my_container` instead of `docker exec`.

Then we trap that signal, execute the python fun and SIGHUP the kopano-server process.

This approach would be more complicated, but is more correct in usage.

See https://github.com/Yelp/dumb-init#session-behavior
So we could have a bash script always running and trapping SIGUSR1 signal (not SIGHUP) that kopano-server ignores at first (Do you know if kopano-server ignores SIGUSR1?).
`dumb-init --rewrite 1:10`  routes this signal to kopano-server and our bash script. Our bash script executes python config configurator script  and SIGHUPs kopano-server afterwards.

So we can still `docker kill --signal=SIGHUP my_container` like standard reload but internally it's becoming SIGUSR1 that becomes SIGHUP again after reconfiguring.

Maybe it's too complicated for the little usage beauty trade.

fbartels:
>Do you know if kopano-server ignores SIGUSR1?

Yes, I do think so. at least this is what my short tests indicate.

I actually havent thought of checking if docker provides an own kill command as well. Since it does I agree that its better to use this instead of running kill through exec.

fbartels:
@zokradonh I have done some more reading about this. do you think we need mapping of signals at all? shouldn't it work to put a trap for HUP into the beginning of the start-services script and then just update the config files as well (maybe this should then move into a function to not duplicate it).

I am not quite sure though if this trapping should still work, since we exec out the process.

rel http://jkamenik.github.io/blog/2017/10/21/docker-details---dumb-init/

zokradonh:
trapping won't work if we exec out.
Thats the reason why I thought of a new script:
>So we could have a bash script always running

So the process tree would look like:
```
dumb-init
-- kopano-server
-- /bin/sh signal-trapper.sh
```

We could tell the users that they can reload configuration by `docker kill --signal=SIGUSR1` and we do not need rewriting. But if we want `docker kill --signal=SIGHUP` we need to make sure, that kopano-server does not get this signal before we have executed the python fun.

dump-init routes the signal to all subprocesses.

fbartels:
closing for now

PR #171, opened 2019-06-07
basis for discussion: make portmapping switchable

cguentherTUChemnitz:
**move portmapping into docker-compose.portmapping.yml and load it automatically**

This changeset enables the possiblity to switch off all host port mappings without patching the docker-compose.yml. Only the COMPOSE_FILE list has to be patched in the .env.

**content of this PR:**
This PR does not change the behavior of this kopano stack. Only the portmapping part is splitted up.

Ensure loading of docker-compose.portmapping.yml via COMPOSE_FILE in .env. This disables the override automatism and starts only when the override.yml exists. This problem is bypassed by enforcing the creation an empty override.yml during setup.sh.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ KOPANO_ZPUSH_REPOSITORY_URL := http://repo.z-hub.io/z-push:/final/Debian_9.0/
 RELEASE_KEY_DOWNLOAD := 0
 DOWNLOAD_COMMUNITY_PACKAGES := 1
 
-COMPOSE_FILE := docker-compose.yml
+DOCKERCOMPOSE_FILE := docker-compose.yml
 TAG_FILE := build.tags
 -include .env
 export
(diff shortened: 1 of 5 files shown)


fbartels:
@cguentherTUChemnitz can you rebase on top of https://github.com/zokradonh/kopano-docker/pull/173 to see if that fixes the travis failure of this pr?

cguentherTUChemnitz:
> @cguentherTUChemnitz can you rebase on top of #173 to see if that fixes the travis failure of this pr?

Thanks for your quick fix!

fbartels:
@zokradonh what do you think?

It feels to me like before merging something like this we would need to think about an update strategy for people just doing a git pull.

cguentherTUChemnitz:
> It feels to me like before merging something like this we would need to think about an update strategy for people just doing a git pull.

This is really a good and non trivial point, to ensure an existing docker-compose.override.yml for those who had run setup.sh earlier.

fbartels:
Not only that, the COMPOSE_FILE also needs to be in .env for ports to stay exposed for existing setups.

cguentherTUChemnitz:
Okay, that seems like a general discussion about an update / upgrade path, where changes in the already generated .env is necessary to fit the upgraded version of the stack. So do i see it right, that the update path via git pull is currently only possible, if the setup.sh would generate the same entries in the global .env file? This does not look very easy, since this would involves patching of the existing config file, which could be modified by the user earlier on.

fbartels:
has been merged in #182

PR #177, opened 2019-06-12
Use jq's -r switch instead of sed

zokradonh:
Use jq's -r switch instead of sed to remove double-quotes.


fbartels (approved):

fbartels:
@zokradonh btw. when I click on the "dockercloud" CI I only get an error 500. Not quite sure what it adds over Travis as well.

zokradonh:
dockercloud CI sets the hub.docker.com image description, thats the reason I have this automated build still on. Normal `docker push` does not set it.

Though, since some month it's possible to set it externally by this:
https://stackoverflow.com/a/55082686/1431178

Maybe we should implement this and disable dockercloud CI completely.

fbartels:
Ah, that is neat. The last time I looked for this the internet still said that this is not possible. Do you want to work on this or shall I give it a try?

The linked github repo also has some other interesting script for deleting old tags, etc. (kopano_base is quite polluted with tags).

PR #181, opened 2019-06-13
wip: add some more documentation about used ports

fbartels:
also add more ports to docker-compose.yml

Found this old commit that I did not yet open a pr with. Taking https://github.com/zokradonh/kopano-docker/pull/171 into account its probably better to have a dedicated compose file per service (that only specifies the exposed ports of this service). 

Also the documentation for ports should move from the core readme to the service in question.


fbartels:
was actually already merged

PR #182, opened 2019-06-14
Make it possible for users to customise exposed ports

fbartels:
My take on #171


fbartels:
Hi @cguentherTUChemnitz,

I have taken your approach and made some adjustments to it. What do you think if this?

Pinging previous contributors for additional feedback: @mithomas @mhoffmann75 @nupplaphil

mhoffmann75:
Although this makes the whole project a bit more complex, I like the idea. I really appreciate the addition to setup.sh for adding missing values. We should keep this up for further changes, too.

However, maybe we should point out preferred way of ignoring docker-compose.portmapping.yml, in the docs, as adding it to .gitignore most probably won't work as it is overridden on next pull... Maybe .git/info/exclude could come to the rescue here ?

fbartels:
Yes, that is a good point with ignoring it. One idea would be to exclude a folder like `~/local-config` and ask users to put their files into this.

cguentherTUChemnitz:
Sorry for the delay, i am currently really packed with work.  Thanks @fbartels for taking my thoughts a step further ;) It seems pretty nice for me with the documentation you already added.

fbartels:
After a bit more thinking the easiest is to just instruct users to put these local port modifications into the override file. instructions have been update. will merge now.

PR #199, opened 2019-07-03
container restart unless stopped

cguentherTUChemnitz:
When i restart my docker host a lot of kopano services do not restart. So i added a lot of restart statements in my override.yml. This works pretty fine, but i think the restart-unless-stopped mechanism could be the default one for this stack. So here is my PR :)

As far as i understand, this hits every service, except the ssl one, because this is only fired up once for initialization. So i omitted the restart statement there.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,6 +44,7 @@ services:
 
   ldap-admin:
     image: osixia/phpldapadmin:0.7.2
+    restart: unless-stopped
     depends_on:
       - ldap
     environment:
@@ -55,6 +56,7 @@ services:
 
   password-self-service:
     image: tiredofit/self-service-password:3.0
+    restart: unless-stopped
     domainname: ${LDAP_DOMAIN}
     depends_on:
       - ldap
@@ -179,6 +181,7 @@ services:
 
   kopano_server:
     image: ${docker_repo:-zokradonh}/kopano_core:${CORE_VERSION:-latest}
+    restart: unless-stopped
     hostname: kopano_server
     container_name: ${COMPOSE_PROJECT_NAME}_server
     depends_on:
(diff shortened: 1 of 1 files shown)


fbartels:
Hi @cguentherTUChemnitz,

thanks for your pr. Especially for the scheduler I had only recently removed this option since I saw repeated starts when the initial startup failed https://github.com/zokradonh/kopano-docker/pull/188/files.

Maybe to have this consistent the value for this service should be switched to an explicit "no"

For the rest of the compose file the restart command seems like a good idea.

docker-compose.yml
```diff
@@ -479,6 +497,7 @@ services:
 
   kopano_scheduler:
     image: ${docker_repo:-zokradonh}/kopano_scheduler:${SCHEDULER_VERSION:-latest}
+    restart: unless-stopped
```
fbartels: As already commented I would rather go with `restart: no` here

fbartels (changes requested):

docker-compose.yml
cguentherTUChemnitz: Ok i know the hassle with endlessly restarting services, when there is some problem on startup. Specifying the `restart: no` will result in a not-running scheduler, when the docker host gets an reboot. So when the scheduler is used to create backups, this should be treated carefully, since this will disable backups as soon as the admin restarts the host.

When i was debugging or fixing not properly starting services, i used to stop them manually. So calling manually `docker-compose stop kopano_scheduler`, will prevent it from automatically restarting. So this usage mimics my real wanted solution of an restart unless stopped or failed, which is sadly not available on the docker daemon.

@fbartels do you think that the risk of host restart disabled backups are treated with enough care, when adding some documentation?

fbartels: I'm sorry, it seems the "no" needs to be quoted.

From experience I know: people don't read documentation ;-) so the only thing that protects you from containers not running, and backups not happening is monitoring, monitoring, monitoring.

fbartels (approved):

PR #209, opened 2019-08-01
Update kweb

### LINKED ISSUE #208: Update kweb to 0.7.0
after updating kweb to the latest release I ran into https://github.com/caddyserver/caddy/issues/2502. 

possible mitigation is to change https://github.com/zokradonh/kopano-docker/blob/3b8c9c4d68cd4c2e4288dfbb34fcd8718440df0f/web/kweb.cfg#L1 and https://github.com/zokradonh/kopano-docker/blob/3b8c9c4d68cd4c2e4288dfbb34fcd8718440df0f/web/kweb.cfg#L5 to *:port instead of :port

fbartels:
fixes #208


web/Dockerfile
```diff
@@ -1,11 +1,18 @@
...
+ARG CODE_VERSION=0.7.0
 FROM kopano/kwebd:${CODE_VERSION}
 
 ARG VCS_REF
 ARG CODE_VERSION
 
 ENV CODE_VERSION="${CODE_VERSION}"
 
+ENV KWEBD_USER root
+ENV KWEBD_GROUP root
```
longsleep: Why run kwebd as root?

web/kweb.cfg
```diff
@@ -1,8 +1,9 @@
-:80 {
+*:80 {
     redir / https://{host}{uri}
 }
 
-*, :443 {
+# kweb >=0.7.0 needs set the explicit hostname (domain needs to be detectable by sni)
+{%FQDN%}, *, *:443 {
```
longsleep: It should not be required to add the FQDN here expliclty.

web/Dockerfile
```diff
@@ -1,11 +1,18 @@
...
+# hadolint ignore=DL3002
+USER root
```
longsleep: This should change to a non root user later on again.

longsleep (changes requested):

web/Dockerfile
fbartels: without `USER root` kweb does not start since its listening on privileged ports: `2019/08/05 11:24:42 Listen: listen tcp :80: bind: permission denied
`

longsleep: Just change internal ports to 8080 and 8443 should do the trick

longsleep: Ah I remember now - not possible because of internal requests right?

fbartels: yes, that was indeed the reason behind this.

longsleep: So @fbartels to fix this set the caps on the kwebd file

`setcap 'cap_net_bind_service=+ep'  /path/to/kwebd`

longsleep: Guess that is something worth adding to the standard kwebd docker image

web/Dockerfile
fbartels: with setcap in the upstream image, this is not necessary any longer

web/kweb.cfg
fbartels: after investigation we came to the conclusion that for lets encrypt cert generation to work the fqdn indeed needs to be included

PR #211, opened 2019-08-07
add example to run owncloud along with Kopano

fbartels:
Open todos:

- [x] configure ldap backend for owncloud
- set displayname and attribute for email correctly in owncloud


cguentherTUChemnitz:
I am just curious, why did you choose owncloud here in favor to nextcloud?

fbartels:
For me it's not much about the "now" of Owncloud, but more about the future (so the upcoming phoenix UI and the infinity scale backend).

On top of this I think from a business perspective its probably the better solution.

owncloud/owncloud.yml
```diff
@@ -0,0 +1,87 @@
...
+    volumes:
+      - oc_files:/mnt/data
+      - ./owncloud/99-ldap.sh:/etc/owncloud.d/99-ldap.sh
+    networks:
+      - web-net
+      - owncloud-net
+      - ldap-net
+
+  oc_db:
+    image: webhippie/mariadb:latest
```
tboerger, line 50: For the sake of simplicity you could use the same database image as on other places, e.g. `mariadb:10.3.10-bionic`.

owncloud/owncloud.yml
```diff
@@ -0,0 +1,87 @@
...
+    restart: always
+    environment:
+      - MARIADB_ROOT_PASSWORD=${MARIADB_ROOT_PASSWORD}
```
tboerger, line 53: In that case you just need to change these env variables

owncloud/owncloud.yml
```diff
@@ -0,0 +1,87 @@
...
+      - MARIADB_USERNAME=${OWNCLOUD_DB_USERNAME}
+      - MARIADB_PASSWORD=${OWNCLOUD_DB_PASSWORD}
+      - MARIADB_DATABASE=owncloud
+      - MARIADB_MAX_ALLOWED_PACKET=128M
+      - MARIADB_INNODB_LOG_FILE_SIZE=64M
+    healthcheck:
```
tboerger, line 59: And also drop this healthcheck as this is not included in the official images.

owncloud/owncloud.yml
```diff
@@ -0,0 +1,87 @@
...
+      timeout: 10s
+      retries: 5
+    volumes:
+      - oc_mysql:/var/lib/mysql
+      - oc_backup:/var/lib/backup
+    networks:
+      - owncloud-net
+
+  oc_redis:
+    image: webhippie/redis:latest
```
tboerger, line 71: Here you could also simply use the official images. I have just taken webhippie images as I'm in control of them :)

tboerger (approved):

owncloud/owncloud.yml
fbartels: Yes, indeed. I wanted to keep as much as possible of your compose file for the moment.

For another pr I have meanwhile implemented a helper script which makes it possible to server multiple databases from the same db server. so this could be used here in the future as well.

PR for reference https://github.com/zokradonh/kopano-docker/pull/169

PR #212, opened 2019-08-09
use the cleaned fqdn inside of kweb.cfg

fbartels:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     restart: unless-stopped
     environment:
       - EMAIL=${EMAIL:-off}
-      - FQDN=${FQDN?err}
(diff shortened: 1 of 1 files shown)


cguentherTUChemnitz:
I assume with this change, we need some other work-around for the self-managed front-facing proxy setup as described here:
https://github.com/zokradonh/kopano-docker/blob/master/examples/traefik-proxy-subdomain.md

Is this part still valid?
```
3. ensure kwmserver is able to connect through an enpoint with valid ssl-certificate
FQDNCLEANED=somethingInvalidToEnforceConnectionFromOutsideEndpoint
```

fbartels:
Yes, I would think that this does not impact the above.

Issue #216, opened 2019-08-15
Multiple webapp plugins trigger errors on build


zokradonh:
Versioning fails if multiple webapp plugins are defined in `.env`.

Reason:
https://github.com/zokradonh/kopano-docker/blob/ae7f679be27c70350489e97a3b9704353d2ce178/setup.sh#L337

This line as it is currently allows use in `Makefile` but fails in `version.sh`:
```
/tmp/kopano-docker-env.BL8re3: line 91: kopano-webapp-plugin-spell-de-de: command not found
```
With enclosed quotes allows use in `version.sh` but then fails in `Makefile`:
```
--build-arg ADDITIONAL_KOPANO_WEBAPP_PLUGINS=""kopano-webapp-plugin-spell kopano-webapp-plugin-spell-de-de kopano-webapp-plugin-spell-en"" \
```

`sh` interprets the quotes but `make` takes them literally on `include`.

fbartels:
Hi @zokradonh,

yes I have already came across the point that the way of handling some values in .env is not really practical when this file is also sourced by some of the bash script. 

the current solution for this is https://github.com/zokradonh/kopano-docker/blob/master/setup-update-tag.sh#L16-L23, but maybe this needs a different handling altogether.

zokradonh:
Ok, so we have three consumers of the `.env` file?

1. Some bash scripts
2. ~~Postfix~~ docker-compose
3. Makefile

fbartels:
You have to replace postfix by docker-compose. It's just primarily the mta image depending on unquoted variables. The quick fix would be moving these values into the dedicated per container env files.

PR #218, opened 2019-08-21
Double quotes not needed by docker compose

zokradonh:
Same issue as https://github.com/zokradonh/kopano-docker/issues/216


fbartels:
@zokradonh can you rebase against master? afterwards the test should continue further than the linting check.

While I am not sure if this really makes a difference (I just tested against ldap, but did not notice something out of the ordinary with it) I am quite sure that it makes a difference for the bash scripting and will then throw the same errors as in the linked ticket.

zokradonh:
This variable is never used in bash, is it?

OpenLDAP is default setting. The kcconf.py script does not work with the double quotes according to my tests. It worked after removing them.

fbartels:
No, not used in any of the bash scripts, but the whole file is sourced which will throw an error (based on experience, have not tested this change though).

> The kcconf.py script does not work with the double quotes according to my tests. It worked after removing them.

Hmm.. Maybe some tests would be in order to check config files after startup.

fbartels:
I have remodeled the package selection and fixed the errors thrown in version.sh

edit: for whatever reason my last local commit is not pushed here.. will take a closer look tomorrow (was wrong remote).

fbartels:
Tests currently fail, since the nightly builds are not available. currently running the tests locally against the supported packages. will merge later if they succeed.

PR #221, opened 2019-08-22
make it possible to redirect to another url, than /webapp

fbartels:


fbartels:
Local test run succeeded

PR #227, opened 2019-08-28
core-start-service: make the grapi ldap-backend usable

bjoernneumann:


fbartels:
@bjoernneumann I did make some changes to be compatible again with the Kopano backend for grapi. Can you do a quick check if it still works as desired for you? (I would guess so, but did not try it)

I currently have a build running to push containers based on your changes to the docker registry.

PR #231, opened 2019-08-30
Add variables to change names of the other docker container

RobinvG:
modified:   Dockerfile
	modified:   kweb.cfg


fbartels:
Hm.. that build does succeed on my system, but consistently fails on travis. will need to dig deeper.

fbartels:
Was able to locally reproduce the issue after clearing volumes and doing a local rebuild. There was a typo in the default env vars in the web container.

Fix has been pushed and tests are currently running @RobinvG .

PR #232, opened 2019-08-30
Repo script

fbartels:
- [x] adapt path for supported archives
- [x] included test no longer fails for pre-final archive


fbartels:
The problem with this is that for example for community tars the url is

https://download.kopano.io/community/core:/core-8.7.83.106.8a1aeef-Debian_10-amd64.tar.gz

while for supported it is 

https://download.kopano.io/supported/core:/final/tarballs/core-8.7.5.0_0%2B43-Debian_9.0-amd64.tar.gz

PR #234, opened 2019-09-05
Added automatic Z-Push shared folder registration.

mithomas:
Via a new environment variable containing the relevant information in a
JSON string, shared folders can be specifified which will be assigned
to all users during container startup.

Z-Push requires manual adding of shared users for each user - for
globally shared folders this is quite tedious. Defining this on service
stack level for automatic setup removes this burden (even though this
requires a container restart).

If the folders are already assigned, nothing happens. Folder removal
is not covered by this.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -210,6 +210,9 @@ services:
       - zpushstates/:/var/lib/z-push/
     environment:
       - TZ=${TZ}
+      # Shared folders automatically assigned to all users in the format: [{"name":"<folder name>","id":"<kopano folder id>","type":"<type>","flags":"<flags>"},...]
+      # For more information on the parameters see the z-push-admin help for the addshared-action.
+      - ZPUSH_ADDITIONAL_FOLDERS=[]
     env_file:
       - kopano_zpush.env
     networks:
--- a/zpush/commander.yaml
+++ b/zpush/commander.yaml
@@ -35,6 +35,41 @@ tests:
     stdout:
       contains:
         - "define('USERNAME', 'SYSTEM');"
+  start-service script (no additional folders):
+    command: bash -c "shopt -s expand_aliases; alias exec='echo exec'; alias php-fpm7.0='echo php-fpm7.0'; . /kopano/start.sh" && cat /etc/z-push/z-push.conf.php
+    exit-code: 0
+    stdout:
+      contains:
+        - "  $additionalFolders = array(\n  );"
+      not-contains: # default entry
+        - "\t$additionalFolders = array("
+        - "\t);"
+  start-service script (empty additional folders):
+    command: bash -c "shopt -s expand_aliases; alias exec='echo exec'; alias php-fpm7.0='echo php-fpm7.0'; . /kopano/start.sh" && cat /etc/z-push/z-push.conf.php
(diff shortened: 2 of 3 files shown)


fbartels:
Hi @mithomas,

thanks for your pr. At the moment the startup of the z-push container fails in the testsuite, but its not directly apparent as to the "why".

But before digging into the startup failure (which could be an undefined var or the added script failing) I must say that I am not sure if this should be a startup action at all. You have to get the folder any anyways during runtime, so why not also run the script to add the folder during runtime (and if you're worried about doing this to newly added users this could be done with the userscripts on user creation).

So in short:
- I like the idea and to automate it (although at the same time I am wary about overusing it, but lets put this aside for the moment)
- I would not do it at container startup
- rather put it as a script in the container
- for new users this could be handled as a userscript

mithomas:
I did notice the failing build and could not figure out why neither.

Anyway, I did just notice that there as actually a way better place to handle this - z-push actually has the $additionalFolders setting for that. So probably `start.sh` should be just set that - what do you think?

fbartels:
Yes, it seems I totally blanked out in regards to the additionalfolders setting as well. Yes, as this is a config value this of total appropriate to set it at container startup.

mithomas:
Next try, this time using $additionalFolders.

This [particular build](https://travis-ci.com/zokradonh/kopano-docker/builds/130590099) errored due to a timeout against ttp://repo.z-hub.io/z-push:/final/Debian_9.0/ when checking the z-push version via `version.sh`.

Since I cannot rerun the build for this PR, I did so on my repository - however, there the [build](https://travis-ci.com/mithomas/kopano-docker/builds/130591516) fails when building the base-image with `invalid reference format` due to a missing repository name in the docker build command: `docker build [...] -t /kopano_base base/` (most likely due to missing docker credentials on my part). Can the build be re-run or do I have to create an artificial commit to trigger it?

Side note: When checking successful logs, I noticed that image size calculation in travis is not working: `The image is  bytes`

fbartels:
Hi @mithomas,

I have meanwhile removed the size information bit from the Makefile. It did work for me locally, but for a non apparent reason it did not work as part of a string on travis. But it was more for general information anyways so I removed it again.

I have yesterday evening merged new code that makes it now easy to perform more detailed tests (I would not go as far as calling it a unit test, but it comes quite close) on the startup scripts. Testing these is then as easy as calling `COMMANDER_OPTS="--concurrent 1 --verbose" COMMANDER_FILES_PATH=zpush dccommander run kopano_zpush` (after installing two dependencies).

I think the program syntax of commander is quite easy to read and understand, do you think you could add two tests (one with the empty default of `ZPUSH_ADDITIONAL_FOLDERS=[]`, one with multiple folders to be configured). If you prefer that, then I could also add these tests to your branch before merging.

mithomas:
Sure, I can do that once I have setup go 1.12 on my Debian Buster system :-).

fbartels:
Oh, you do not need go. there are already precompiled binaries in the projects repository.

mithomas:
Added commander test cases.

fbartels:
Very nice. Thanks!

Just reran the test and all works as expected.

PR #242, opened 2019-09-15
WIP: separate sockets and pids

### LINKED ISSUE #226: configure sockets to use new mount
currently pids and sockets are mounted into a volume. it would be better to leave pids on tmpfs and have the sockets separately

fbartels:
fixes #226

Todos:
- [ ] scheduler give `The server is not running, or not accessible through "default:".` when trying to create store (needs admin.cfg for new socket)
- [ ] bump versions
- https://stash.kopano.io/projects/KC/repos/kopanocore/pull-requests/3301/overview will remove pids alltogether

we can also define the socket through KOPANO_SOCKET


fbartels:
Overall seems more complicated than useful

PR #243, opened 2019-09-17
Enhance description with relation between containers

fbartels:
Better describe relation between containers by including an architecture description.


fbartels:
- Have a nice graph in draw.io
- unsure if container description should be in one document or in each individual readme
- is there a way to include these information e.g. in the Dockerfile and compile them into a central doc?

PR #253, opened 2019-10-07
update to Debian Buster

fbartels:
Wait for https://github.com/zokradonh/kopano-docker/pull/232

Waiting for a first Kopano release that officially supports Debian 10


docker-compose.yml
```diff
@@ -140,6 +140,7 @@ services:
       - kopano_ssl
       - kopano_konnect
     environment:
+      - DEBUG=true
```
fbartels: needs to be removed again

docker-compose.yml
```diff
@@ -397,6 +398,7 @@ services:
       - kopanossl/:/kopano/ssl
       - kopanosocket/:/run/kopano
     environment:
+      - DEBUG=true
```
fbartels: needs to be removed again

PR #270, opened 2019-10-23
further goss checks

fbartels:
- [x] can goss check the user a process is running with? 
    - seems like such functionality is not included https://github.com/aelsabbahy/goss/issues/338


fbartels:
The testsuite knowingly fails at the moment as it needs a build containing https://jira.kopano.io/browse/KC-1620 to properly startup.

https://jira.kopano.io/browse/KC-1616

PR #273, opened 2019-10-29
Make sure that services listen globally

fbartels:
they need to be made available through reverse proxy and port forwarding anyways
changed in https://jira.kopano.io/browse/KC-1616


fbartels:

PR #274, opened 2019-10-29
update konnect, kwmserver and kweb to the latest version

fbartels:


fbartels:

PR #278, opened 2019-11-03
Fixed issues with Z-Push shared folder configuration.

mithomas:
setup.sh generated wrong config key for folder id and wrong datatypes
for folder type and flags. This has been corrected both there and in
the automated tests.

Automated tests did not catch it since they suffered the same issues; manual integration testing did not catch it since the respective folders in my setup were already registered manually, a fact that I missed. The error popped up after checking the log file, which complained about it.

### DIFF
--- a/zpush/commander.yaml
+++ b/zpush/commander.yaml
@@ -61,7 +61,7 @@ tests:
     exit-code: 0
     stdout:
       contains:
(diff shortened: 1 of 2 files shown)


mithomas:
Hm - failing build seems to show the same issue as build [#928](https://travis-ci.com/zokradonh/kopano-docker/builds/134777061) from `master` and unrelated to my changes. Once this is addressed I'll rebase if needed.

fbartels:
Hi @mithomas,

thanks for your pr. I am not quite sure where setup.sh comes into this though, do you mean start.sh?

I have started a rebuild of your pr, it seems the deadlock database issue is not entirely resolved. As this seems like a timing issue its hard to get behind this though (I thought this was resolved with creating the database before server startup, at least it did not happen for quite a while afterwards).

mithomas:
Yeah, `start.sh` of course - I have amended the commit message.

fbartels:
Tests are green after rebuild. Merging.

PR #295, opened 2019-11-26
Disable some checks when deploying with injected cfg files

RobinvG:
With Kubernetes you can use configmaps to inject the cfg files, therefore you don't need any environment variables 

## Proposed Changes
  - Add a check that prevents waiting on files that depend on environment variables
  - Continue if a config file is read only when changing options
  - Add a check that Ignores the config parser completely

### DIFF
--- a/base/kcconf.py
+++ b/base/kcconf.py
@@ -4,6 +4,7 @@ via environment variables"""
 import re
 import os
 import os.path
+import sys
 
 def configkopano(configs):
     """ Changes configuration files according to configs typically returned from parseenvironmentvariables(..)"""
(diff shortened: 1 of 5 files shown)


fbartels:
Hi Robin,

I don't think the python enhancement already works as expected. I have added some test cases (which still need a tweak to my commander wrapper), but the check with an immutable server.cfg fails with the following:

```
Traceback (most recent call last):
  File "/kopano/server.py", line 16, in <module>
    'kcoidc_initialize_timeout': "360"
  File "/kopano/kcconf.py", line 37, in configkopano
    with open(filename, "w") as f:
PermissionError: [Errno 13] Permission denied: '/etc/kopano/server.cfg'
```

edit: on second thought that could have been because i only did a rebuild of the core image, and not the base image.

PR #297, opened 2019-11-27
Use variables to check if the server is running

RobinvG:
Add server_socket variables so we can use another name then kopano_server as the server name

### DIFF
--- a/core/start-service.sh
+++ b/core/start-service.sh
@@ -4,6 +4,13 @@ ADDITIONAL_KOPANO_PACKAGES=${ADDITIONAL_KOPANO_PACKAGES:-""}
 KCCONF_SERVER_MYSQL_SOCKET=${KCCONF_SERVER_MYSQL_SOCKET:-""}
 DISABLE_CHECKS=${DISABLE_CHECKS:-false}
 DISABLE_CONFIG_CHANGES=${DISABLE_CONFIG_CHANGES:-false}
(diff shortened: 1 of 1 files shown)


fbartels:
Hi @RobinvG,

I was wondering if there is a way to simplify the environment variables until I realised that this way the same env variable is used in the configuration file for the socket this way. I like that.

The downside is that this does not seem to work for dockerize:

```
kopano_gateway_1    | 2019-11-27T15:58:39.422520256Z 2019/11/27 15:58:39 Waiting for http://kopano_server:236/: unexpected HTTP status code: 405.
kopano_ical_1       | 2019-11-27T15:58:20.360662763Z 2019/11/27 15:58:20 Waiting for http://kopano_server:236/: unexpected HTTP status code: 405.
```

fbartels:
Maybe this needs a single "KOPANO_CONN" after all. Then there is one place where one can check if the variable begins with with "http", replace "http*://" with "tcp://" and finally assign this value to KCCONF_*_SERVER_SOCKET.

fbartels:
@RobinvG what do you think about https://github.com/zokradonh/kopano-docker/pull/298 ?

I think this approach may be less confusing down the road.

RobinvG:
@fbartels  I agree let's use that one then

fbartels:
Great. Currently waiting for tests to finish, then I'll merge it.

PR #303, opened 2019-12-10
Prepare usage of build secrets

### LINKED ISSUE #75: Can we use apt_auth to more securely store repo credentials?
https://manpages.debian.org/testing/apt/apt_auth.conf.5.en.html

apt_auth.conf could be removed in the same layer it was created to not accidentally store credentials in the resulting image. Having credentials in a separate file also prevents us from removing the complete repo information (so the containers could still be able to install additional packages at startup).

startup script would need to be modified to add the credentials back at runtime.

fbartels:
- sort .gitignore and add apt_auth.conf to it
- add helper to setup.sh to create apt_auth.conf (user still needs to enter their own credentials)
- add tooling to core dockerfile and makefile to use build secrets
- fix build on travis
- [x] update readme
  - [x] document minimal version for building
  - [x] update builder container
- [x] implement secrets for webapp

fixes https://github.com/zokradonh/kopano-docker/issues/75

should be rebased and merged after https://github.com/zokradonh/kopano-docker/pull/253


fbartels:
Maybe this could finally be the reason to completely move to Github Actions?

Issue #310, opened 2019-12-19
All containers should run read-only


fbartels:
To easily detect unexpected writes the containers should run read-only themselves.

Relates to https://github.com/zokradonh/kopano-docker/issues/309

PR #314, opened 2020-01-02
Run all containers read-only

### LINKED ISSUE #309: Prevent modifications to mounted files
All files that are modified during container startup should be copied into a temp location before modifying them.

For example:
https://github.com/zokradonh/kopano-docker/blob/b36093d0743587247e7ce8418c6f346b3a094731/konnect/wrapper.sh#L81-L85

When these files are mounted into the container modification could fail otherwise.

fbartels:
Fixes #309 and #310


core/start-service.sh
```diff
@@ -31,8 +35,9 @@ ADDITIONAL_KOPANO_PACKAGES=$(echo "$ADDITIONAL_KOPANO_PACKAGES" | tr -d '"')
 	fi
 done
 
-mkdir -p /kopano/data/attachments /kopano/data/kapi-kvs /tmp/"$SERVICE_TO_START" /var/run/kopano /var/lib/kopano-grapi
+mkdir -p /tmp/"$SERVICE_TO_START" /var/run/kopano
 
+# TODO is this still required now that we won't modify configuration mounted to /etc/kopano?
```
fbartels, line 42: @RobinvG  with this PR config files are moved to /tmp before being updated. This makes it seem like the below if is not required anymore. What do you think?

PR #331, opened 2020-02-03
Add wrapper scripts to easily get admin.cfg

### LINKED ISSUE #330: how to correctly set the langauge in .env so I don't need to modify the docker itself.
I get the problem that new users get an english store.
I set the entry "MAILBOXLANG=de_DE.UTF-8" inside ".env"

changing the store manually using "kopano-storeadm -D -n username@domain.tld && kopano-storeadm -C -l de_DE.UTF-8 -n username@domain.tld" works.

fbartels:
this pr adds wrapper scripts for kopano-admin, kopano-cli and kopano-storeadm to make sure that when these commands are invoked that always use the dynamically created admin.cfg in /tmp/kopano.

Fixed #330 

Additionally this adds some workaround code to no longer let tests fail because kopano-admin --sync fails to create stores.

### DIFF
--- a/core/bin/TODO
+++ b/core/bin/TODO
@@ -1,0 +1,2 @@
+override create user scripts with own?
+override kopano-admin binary with own script wrapper to source config from /tmp/kopano?
--- a/core/bin/TODO
+++ b/core/bin/TODO
@@ -1,2 +1,0 @@
-override create user scripts with own?
-override kopano-admin binary with own script wrapper to source config from /tmp/kopano?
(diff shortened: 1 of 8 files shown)


fbartels:
In the end I went with the second option and added wrapper scripts. The change has also been merged for quite a while already.

This of course does not prevent you from adding further scripts to `createuser.d`. You'd only need to mount the at runtime or back them into your own image that is based on this one.

If you think you have a script that would be beneficial for other users then pull requests are welcome!

PR #353, opened 2020-03-05
move implicit defaults to docker-compose

### LINKED ISSUE #294: Discussion: How to make it more transparent what options can be set and which are the default?
There are currently multiple ways how "default" values for configuration values are defined:

- the programmed default of the binary
  - e.g. kopano-servers default for `user_plugin` is `db` (see https://stash.kopano.io/projects/KC/repos/kopanocore/browse/installer/linux/server.cfg#342)
- a "default" derived from the installed configuration file
  - if a file is mounted into the container this could override settings made
- a runtime default enforced at startup
  - e.g. through https://github.com/zokradonh/kopano-docker/blob/master/core/defaultconfigs/server.py
  - this is intended for settings that are expected within this container deployment, but it could differ from what users could experience from a packaged installation
- a default from docker-compose
  - docker-compose specifies a lot of environment variables
  - my primary intention so far was that they are more visible in the compose file and can/should be settings that admins want to modify (plus if you have something variable this is nice as well)

fbartels:
some non-default configuration options are backed into the startup scripts. This can be unexpected when mounting files e.g. in a Kubernetes setup.

Fixes https://github.com/zokradonh/kopano-docker/issues/294

CC @RobinvG


fbartels:
@RobinvG can you have a look at the changes. Since this also requires changes to the meet compose file I am wary that this could cause issues for users of the meet example deployment.

core/defaultconfigs/dagent.py
```diff
@@ -1,14 +1,12 @@
-import os
+#!/usr/bin/env python3
```
longsleep, line 1: What sense does it make to have these python files with shebang?

fbartels: Its a script that is run, so my reasoning was the other way around. Why leave it out?
https://github.com/zokradonh/kopano-docker/blob/master/core/start-service.sh#L52

longsleep: Sure, but please either use a shebang or call it as a script with interpreter. Now its duplicated.

PR #358, opened 2020-03-10
add user with non-ascii username

fbartels:


ldap_demo/bootstrap/ldif/demo-users.ldif
```diff
@@ -4337,3 +4337,24 @@ sn: meetingroom3
 uidNumber: 20055
 uid: meetingroom3
 kopanoUserServer: kopano_server
+
+dn: uid=user24,ou=users,ou=CompanyB,{{ LDAP_BASE_DN }}
```
longsleep, line 4341: And when exporting this, the dn is also base64 encoded and contains the field value?

fbartels: when importing the ldif with phpldapadmin yes, but the import in the ldap image seems to behave differently. here it uses the ascii dn and duplicates the uid value (login will works with the non-ascii variant).

This does not seem to solve the problem, it should solve.

PR #361, opened 2020-03-11
Allow users to use different saslauthd-options.

theoneandonly-vector:


fbartels:
Hi @theoneandonly-vector,

thanks for your pr. I appreciate going the  extra mile and making it even configurable. Since by you other issues the ldap auth does not work I would rather recommend to remove it and just use the auth through imap.

PR #386, opened 2020-04-28
make it possible to specify existing ssl certificate

fbartels:


fbartels:
No, this does not yet have the desired effect (it just introduces a new variable)

PR #393, opened 2020-05-02
add ldap-user-manager

fbartels:
login at /setup still fails
also first tests running it at a subdirectory failed


fbartels:
Closing for the moment as the application does not allow running in a sub directory. Although it does have some nice sides to it.

PR #414, opened 2020-06-14
Fix core gateway start parameter

engelant:
gateway ignored generated config in /tmp/kopano/gateway.cfg and took it's default /etc/kopano/gateway.cfg

Fixes #413


fbartels:
Hi @engelant,

passing the configuration option as an explicit parameter is not necessary, since the config file is already specified through `KOPANO_CONFIG_PATH`.

Up until https://github.com/zokradonh/kopano-docker/pull/352 the config was specified as a parameter.

PR #419, opened 2020-06-24
add kweb config for z-push

fbartels:
relates to #36


zpush/kweb.cfg
```diff
@@ -0,0 +1,23 @@
...
+	rewrite /Microsoft-Server-ActiveSync /Microsoft-Server-ActiveSync/index.php
+	fastcgi2 /Microsoft-Server-ActiveSync /run/php/php7.0-fpm.sock php {
+		without /Microsoft-Server-ActiveSync/
+		root /usr/share/z-push/
+		read_timeout 3605s
+	}
+
+	# https://stash.kopano.io/projects/KGOL/repos/kweb/pull-requests/29/diff#test/Caddfile.test
+	# zpush autodiscover (test!)
+	rewrite /Autodiscover/Autodiscover.xml /Autodiscover/autodiscover.php
```
longsleep: This rewrite needs to be case insensitive.

zpush/kweb.cfg
```diff
@@ -0,0 +1,23 @@
...
+	errors stderr
+
+	# healthcheck
+	status 200 /status
+
+	rewrite /Microsoft-Server-ActiveSync /Microsoft-Server-ActiveSync/index.php
+	fastcgi2 /Microsoft-Server-ActiveSync /run/php/php7.0-fpm.sock php {
+		without /Microsoft-Server-ActiveSync/
+		root /usr/share/z-push/
+		read_timeout 3605s
```
longsleep, line 13: The exact value should be exposed as configuration, probably also could be auto computed from the corresponding z-push configuration value at startup. (just a little higher than the max z-push ping time).

zpush/kweb.cfg
```diff
@@ -0,0 +1,23 @@
+:9080 {
+	log stdout
+	errors stderr
+
+	# healthcheck
+	status 200 /status
+
+	rewrite /Microsoft-Server-ActiveSync /Microsoft-Server-ActiveSync/index.php
+	fastcgi2 /Microsoft-Server-ActiveSync /run/php/php7.0-fpm.sock php {
```
longsleep: Do we really use php7.0 here - if so its time to update the base platform since PHP 7.0 is EOL since a couple of years.

longsleep: Imo we should run with PHP 7.4 (or maybe 7.3 but that will be EOL this december as well).

longsleep (changes requested):

zpush/kweb.cfg
fbartels: No, so far this is just a simple copy of the example you provided. The containers currently use php 7.3 (which is the default provided by Debian 10)

longsleep: Ah ok

PR #440, opened 2020-09-10
Allow to install extra packages

peterfromthehill:
The string in line #20 is always zero, we should watch the return value of mkdir.

### DIFF
--- a/core/start-service.sh
+++ b/core/start-service.sh
@@ -32,7 +32,7 @@ if [ "${AUTOCONFIGURE}" == true ]; then
 	# Hint: this is not compatible with a read-only container.
 	# The general recommendation is to already build a container that has all required packages installed.
 	ADDITIONAL_KOPANO_PACKAGES=$(echo "$ADDITIONAL_KOPANO_PACKAGES" | tr -d '"')
-	if [ -n "$(mkdir -p "/var/lib/apt/lists/" 2&> /dev/null)" ]; then
+	if mkdir -p "/var/lib/apt/lists/" 2&> /dev/null; then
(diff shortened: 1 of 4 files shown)


fbartels:
Hi @peterfromthehill,

thanks for your PR. That check broke when upgrading the base to Debian Buster. There are however a few more places where the code needs to be adapted. Can you include them in your pr?

https://github.com/zokradonh/kopano-docker/blob/d514ef44cb7278210bd10854737d5d03e75a8382/kdav/start.sh#L17
https://github.com/zokradonh/kopano-docker/blob/d514ef44cb7278210bd10854737d5d03e75a8382/zpush/start.sh#L46
https://github.com/zokradonh/kopano-docker/blob/d514ef44cb7278210bd10854737d5d03e75a8382/core/start-service.sh#L35

PS: In general I would recommend to rather build a container that has the additional packages built in. An example for a Dockerfile to do this can be found at https://github.com/zokradonh/kopano-docker/blob/d514ef44cb7278210bd10854737d5d03e75a8382/webapp/Dockerfile.plugins. Including the packages directly in the container image saves you from unexpected behaviour if packages have been updated in between restarts of the container.

fbartels (changes requested):
see above

fbartels:
@peterfromthehill can you give an indication when you could update your pr? If you are not able to do it this week I would make the relevant changes myself in your pr.

peterfromthehill:
@fbartels: Done :). Sorry for the delay, i had some off days.

fbartels:
@peterfromthehill thanks for your changes. And no need to be sorry, just wanted to close the pr and was therefore checking in.

PR #448, opened 2020-09-21
update docker-mailserver to 7.0.0

### LINKED ISSUE #445: Updating mailserver to 7.x
I wanted to ask if anybody attempted to upgrade the mailserver to 7.0 and if there are any known issues.

Reason for that is that there are a few changes made (e.g. a user-script.sh) which can be configured to change (e.g. sed) the generated config.

This is an issue, if you use a single domain authenthication setup, since saslauthd_rimap (part of mailserver) is configured with "-r" by default, which means the username will either be passed trough, if it contains an @ (eg. user@example.com) or appended with the hosts own domain (eg user => user@mail01.example.com). This will lead to the Samba4/LDAP auth to fail for smtp (submission to be precise), since the server expects "user" as the username, not user@mail01.example.com.
See [this issue](https://github.com/tomav/docker-mailserver/issues/1608).
Works fine for imap though, since only postfix adds the domain before attempting to auth against imap.

Also today I discovered  the saslauthd is logging quiet a lot, including the users password. I do not like that.

I hardly think the maintainers of the mailserver will backport anything to the 6.2.0 release, so...

reneploetz:
Fixes #445 

## Proposed Changes
  - updated docker-mailserver image to 7.0.0
  - added volume for mail logs
  - use SPAMASSASSIN_SPAM_TO_INBOX to deliver spam to INBOX, upstreams new default is bouncing: https://github.com/tomav/docker-mailserver#spamassassin_spam_to_inbox
 - migrate REPORT_RECIPIENT to the new PFLOGSUMM_TRIGGER=logrotate

## Notes
- all REPORT_* options have been deprecated upstream and were split into PFLOGSUMM_* and LOGWATCH_* to allow individual configuration between pflogsum (postfix reports) and logwatch mail reports. The latter reports are new and disabled per default. I opted for migrating the options to PFLOGSUMM_* only to have the same behaviour as before. One might consider to enable logwatch reporting, too.
- the PFLOGSUMM_TRIGGER option does allow a new option "daily_cron" which creates report on a daily basis instead of a logrotating basis. This might be more accurate for some users, but I opted to keep the old behaviour which is encoded as "logrotate"
- we should consider setting POSTFIX_INET_PROTOCOLS to "ipv4" to allow running this container on non-ipv6-enabled hosts in a more secure manner: https://github.com/tomav/docker-mailserver/issues/1405#issuecomment-590106498 or https://github.com/tomav/docker-mailserver#permit_docker I cannot test the issue myself as my hosting provider is IPv4 only.

### DIFF
--- a/docker-compose.mail.yml
+++ b/docker-compose.mail.yml
@@ -2,7 +2,7 @@ version: "3.5"
 
 services:
   mail:
-    image: tvial/docker-mailserver:release-v7.0.0
+    image: tvial/docker-mailserver:release-v7.0.1
     restart: unless-stopped
     hostname: mail # hostname and domainname may need to be commented on some platforms (e.g. ChromeOS)
     domainname: ${LDAP_DOMAIN}
--- a/docker-compose.mail.yml
+++ b/docker-compose.mail.yml
@@ -2,7 +2,7 @@ version: "3.5"
 
 services:
   mail:
-    image: tvial/docker-mailserver:release-v6.2.0
+    image: tvial/docker-mailserver:release-v7.0.0
     restart: unless-stopped
     hostname: mail # hostname and domainname may need to be commented on some platforms (e.g. ChromeOS)
     domainname: ${LDAP_DOMAIN}
@@ -14,6 +14,7 @@ services:
     volumes:
       - maildata:/var/mail
       - mailstate:/var/mail-state
+      - maillogs:/var/log/mail
       - mtaconfig:/tmp/docker-mailserver/
     environment:
       - DMS_DEBUG=0
@@ -35,8 +36,8 @@ services:
       - ONE_DIR=1
       - PERMIT_DOCKER=connected-networks
       - POSTFIX_DAGENT=lmtp:kopano_dagent:2003
+      - PFLOGSUMM_TRIGGER=logrotate
       - POSTMASTER_ADDRESS=${POSTMASTER_ADDRESS}
-      - REPORT_RECIPIENT=1
       - SASLAUTHD_LDAP_BIND_DN=${LDAP_BIND_DN}
       - SASLAUTHD_LDAP_FILTER=${SASLAUTHD_LDAP_FILTER}
       - SASLAUTHD_LDAP_PASSWORD=${LDAP_BIND_PW}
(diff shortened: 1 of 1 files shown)


docker-compose.mail.yml
```diff
@@ -2,7 +2,7 @@ version: "3.5"
 
 services:
   mail:
-    image: tvial/docker-mailserver:release-v6.2.0
+    image: tvial/docker-mailserver:release-v7.0.0
```
fbartels: There seems to be a 7.0.1 already https://hub.docker.com/layers/tvial/docker-mailserver/release-v7.0.1/images/sha256-8d47cb0d2ecc31be1ffa42c5ae988ca61754099c3cc6cc1d8a5f45601e0c592d?context=explore

fbartels (commented):
Hi @reneploetz,

thank you very much for your pr. I have just a small comment which i made inline.

Not too sure about your IPv4 comment. Afair the Docker network is IPv4 only anyways so I don't think this setting will have any real influence.

docker-compose.mail.yml
reneploetz: That's correct. I'll update the image to 7.0.1.

reneploetz:
> Hi @reneploetz,
> 
> thank you very much for your pr. I have just a small comment which i made inline.
> 
> Not too sure about your IPv4 comment. Afair the Docker network is IPv4 only anyways so I don't think this setting will have any real influence.

I read the whole discussion in the upstream ticket again and I think you are right that POSTFIX_INET_PROTOCOLS will not help, but the underlying problem - users might accidently create an open relay - might still be present. I'm unable to verify this tough.


Based on the upstream discussion, the conditions are:
- IPv6 is enabled on the host
- Docker is running in IPv4 only mode (which is the default)
- a port is bound (exposed) to all interfaces (default of this container)

The problem appears to be that if you connect to the host system using the IPv6 address, then Docker will translate this address to IPv4 to allow external systems to connect to the container. The problem is that the translation is working like a NAT - Docker is replacing the source address (which is the external IP) with the IP of the Docker gateway. The gateway server might be a trusted network for postfix tough - so it might be allowed to sent mails without authentication and thus creating an open relay server if you connect via IPv6.

To mitigate this, the upstream tickets suggests to either enable IPv6 in docker or bind to IPv4 only as stated here: https://github.com/tomav/docker-mailserver/issues/1405#issuecomment-590115647
Upstream decided to post a warning for the PERMIT_DOCKER setting to make users aware of the issue.

But I feel that this discussion is probably better done in a separate ticket anyway. So I would suggest to merge this first and maybe think about what to do afterwards if you feel that there is something that should be done.

fbartels:
Thanks again. Your pr has just been merged.

PR #462, opened 2020-09-29
'Repair' ODT filepreviewer

peterfromthehill:
Hi,

when i add the kopano-webapp-plugin-filepreviewer i was normaly able to view odt files inside the webapp. With the caddy/kwebd webserver i wasnt able to view odt files.
If the caddy webserver should serve a index.htm(l) file, its redirect to the directory thats hold the index.htm(l) file.
But with the fastcgi2 or alias module from kwebd this does not work, its generate a false redirect (pls see the issue in kweb github repo).
I change the config to use only buildin caddy modules and get the webapp+plugins to work.

Do you see any breaks in my config?

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/webapp/kweb.cfg
+++ b/webapp/kweb.cfg
@@ -1,31 +1,29 @@
-:9080 {
-	log stdout
-	errors stderr
+:9080/webapp {
+    log stdout
+    errors stderr
 
-	# healthcheck
-	status 200 /status
+    root /usr/share/kopano-webapp/
+    fastcgi / /run/php/php-fpm-kopano-webapp.sock php {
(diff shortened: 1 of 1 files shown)


fbartels:
Hi @peterfromthehill,

thanks for your pr. It currently fails already at the linting stage however since you switched tabs with spaces.

I am in parallel checking if this is indeed the best approach.

fbartels:
As stated in https://github.com/Kopano-dev/kweb/issues/3 the issue will automatically disappear once 0.12.3 is included in the core:master repository.

Issue #398, opened 2020-05-15
Add option for web container to extend csp


fbartels:
Apps such as Meet and Konnect set strict csp that disallow embedding it from other domains.

The web container could extend these to allow embedding from a domain that is specified through an environment variable.

https://caddyserver.com/v1/docs/header

fbartels:
https://stash.kopano.io/projects/KGOL/repos/kweb/browse/snippets.d/60-kopano-staticpwa-csp.cfg

PR #469, opened 2020-10-12
add extra config snippet to override default csp configuration

fbartels:
references https://github.com/zokradonh/kopano-docker/issues/398

but so far the value for frame-ancestors is not appended


fbartels:
Needs to overwrite https://stash.kopano.io/projects/KGOL/repos/kweb/browse/snippets.d/60-kopano-staticpwa-csp.cfg inside of the kopano-meet container

PR #478, opened 2021-03-03
add tmpfs for /var/lib/kopano-webapp/plugin_files

fbartels:


fbartels:
Merged through https://github.com/Kopano-dev/kopano-docker/pull/18

PR #486, opened 2021-09-01
Fixes z-hub repo

mhoffmann75:
Since the z-hub repo has moved we need to change all occurrences of the old repo or the docker containers won't built anymore.
For details see https://kopano.com/z-push/z-push-new-urls-for-repositories-and-wiki/


fbartels:
Hi @mhoffmann75,

thanks for the pr! With this change the z-push image indeed builds again.

PR #493, opened 2022-06-14
fixed volume references, no trailing slash

robinrosenberger:
docker-compose v2.6.0 fails to spin up because of trailing slashes on volume references


fbartels:
Hi @robinrosenberger,

thanks for your pr.

